PR #22, opened 2017-01-07
Bugfixes/schoen/patch misc

mtschoen-unity:
This is some stuff that I fixed directly on the patch branch. I actually didn't fix the issue with HierarchyListViewController scrolling to the selected in the patch, but it's fixed here.

I also noticed an issue with the HighlightModule after Dylan's expand-fix, and fixed it.


amirebrahimi:
Can you merge up latest development?

mtschoen-unity:
This should be up to date. Merge commit is 	33adac8

Workspaces/Hierarchy/Prefabs/HierarchyListItem.prefab
```diff
@@ -216,7 +216,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_RunInEditMode: 0
-  m_SelectionFlags: 2
+  m_SelectionFlags: 3
```
amirebrahimi: Why this change?

mtschoen-unity: Consistency. Timoni (and Dennis) have asked that we switch everything to ray-and-direct "where possible." This is for the expand-arrow, which we weren't actually seeing until I changed its material to render on top of the list items.

Scripts/Modules/Highlight/HighlightModule.cs
```diff
@@ -17,11 +19,10 @@ void LateUpdate()
 			if (go == null)
 				continue;
 
-			var highlightTransform = go.transform;
-			var highlightScaleIncrease = Vector3.one * 0.0125f;
-			Matrix4x4 highlightScaleIncreaseMatrix = Matrix4x4.TRS(highlightTransform.position, highlightTransform.rotation, highlightTransform.lossyScale + highlightScaleIncrease);
 			foreach (var m in go.GetComponentsInChildren<MeshFilter>())
 			{
+				var highlightTransform = m.transform;
```
mtschoen-unity: Yes. It was creating the transformation matrix once for the parent object, thus it was drawing all objects in the same position/rotation/scale. This moves the TRS call into the foreach loop, and turns the scale increase into a static/readonly field which makes it a compile-time constant. Not sure if this makes a huge difference but stylistically I like to have those tweakable hardcoded values defined in the class scope.

Scripts/Modules/Highlight/HighlightModule.cs
```diff
@@ -5,6 +5,8 @@
 
 public class HighlightModule : MonoBehaviour
 {
+	static readonly Vector3 kHighlightScaleIncrease = Vector3.one* 0.0125f;
```
amirebrahimi: space after Vector3.one

mtschoen-unity: Well hot damn... we can make changes directly in GitHub. I call that a win.

PR #27, opened 2017-01-11
Features/schoen/integration testing

mtschoen-unity:
This PR adds a few integration tests.

TestCCU - checks for the UNITY_CCU define. I made the const string public to avoid copy/pasting it. Because the CCU class and namespace shared the same name, I could not use a #using directive. I renamed the namespace to just ConditionalCompilation.

NoDependencyTest - Attempts a compile using a custom set of #defines, specifically excluding UNITY_CCU and the defines it adds. If there are errors, the test fails.

BuildPlayerTest(s) - Attempt a build with VR-capable players, fail on errors. If the player is not installed, the result is Inconclusive. For automated integration testing, we should either omit the non-standalone player tests, or use a Unity build with all players installed

TagManagerTest(s) - Check for required tags and layers, fail if any are missing. I ended up extracting GetRequiredTags/Layers and GetTagManagerAsset into public static methods in order to avoid duplicating these procedures. I assumed that tests were a case where the "Don't reach in" rule doesn't apply.

This PR addresses issue #29


mtschoen-unity:
Quick update on the situation with CloudBuild. We need to add an empty Assets and ProjectSettings folder or else UCB aborts because it doesn't think this is a project. The ucb_prebiuld.sh script moves the whole repo into the Assets folder.

We are now waiting on the UCB team to get our preview build set up as a Unity version to build against. Once this is done, we should be able to run tests.

Since the empty Assets folder is probably not a permanent solution, we should work with the UCB team to come up with a better solution. I probably should have branched those changes off since the tests themselves are still useful and could be merged back. If we don't figure out a better way to automate the tests, we can just delete Assets/ ProjectSettings/ and ucb_prebiuld.sh after we review.

Tests/Editor/BuildPlayerTest.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		[Test]
+		public void TestStandaloneLinux()
+		{
+			TestBuildPlayer(BuildTarget.StandaloneLinux);
+		}
+
+		[Test]
+		public void TestStandaloneLinux64()
```
amirebrahimi: Remove

Tests/Editor/BuildPlayerTest.cs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		[Test]
+		public void TestWebGL()
+		{
+			TestBuildPlayer(BuildTarget.WebGL);
+		}
+
+		[Test]
+		public void TestStandaloneLinux()
```
amirebrahimi: Remove

Tests/Editor/BuildPlayerTest.cs
```diff
@@ -0,0 +1,91 @@
...
+	{
+		const string kTestPath = "_Test";
+		const string kVerifySavingAssets = "VerifySavingAssets";
+
+		bool? m_VerifySavingAssets;
+
+		[SetUp]
+		public void SetUp()
+		{
+			if (EditorPrefs.HasKey(kVerifySavingAssets))
```
amirebrahimi: Need better workflow for this. If I run tests one-by-one, then I get the prompt each time.

Scripts/Utilities/Editor/ConditionalCompilationUtility.cs
```diff
@@ -30,9 +30,9 @@ namespace ConditionalCompilationUtility
 	/// }
 	/// </summary>
 	[InitializeOnLoad]
-	public class ConditionalCompilationUtility
+	public static class ConditionalCompilationUtility
 	{
-		const string kEnableCCU = "UNITY_CCU";
+		public const string kEnableCCU = "UNITY_CCU";
```
amirebrahimi: Instead of making this public create a static bool property enabled that does the check you do in the test.

Tests/Editor/CCUTest.cs
```diff
@@ -0,0 +1,86 @@
...
+				Path.Combine(Path.Combine(Path.Combine(Path.Combine(extensionsPath, "Unity"), "EditorTestsRunner"), "Editor"), "UnityEditor.EditorTestsRunner.dll")
+			};
+
+			//GetAllFiles(EditorApplication.applicationContentsPath, references, "*.dll"); // Need to weed out unmanaged dlls
+			//GetAllFiles(dependencyPath, references, "*.dll"); // Error on loading ICsharpCode.NRefactory
+
+			var sources = new List<string>();
+			GetAllFiles(Application.dataPath, sources, "*.cs");
+
+			// TODO: Find a better way to collect #defines
```
amirebrahimi: defines = EditorUserBuildSettings.activeScriptCompilationDefines - defines that come from CCU (create a separate property or method for those)

Tests/Editor/CCUTest.cs
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		[Test]
+		public void NoDependencyTest()
+		{
+			// TODO: Find a better way to collect dependencies
+			var dependencyPath = Path.Combine(EditorApplication.applicationContentsPath, "Managed");
+			var extensionsPath = Path.Combine(EditorApplication.applicationContentsPath, "UnityExtensions");
+
+			var references = new List<string>
```
amirebrahimi: Use http://stackoverflow.com/questions/383686/how-do-you-loop-through-currently-loaded-assemblies/26300241#26300241 + http://stackoverflow.com/questions/52797/how-do-i-get-the-path-of-the-assembly-the-code-is-in

Tests/Editor/CCUTest.cs
```diff
@@ -0,0 +1,86 @@
...
+			GetAllFiles(Application.dataPath, sources, "*.cs");
+
+			// TODO: Find a better way to collect #defines
+			var defines = new[]
+			{
+				"UNITY_EDITOR",
+				"UNITY_5_3_OR_NEWER"
+			};
+
+			const string outputFile = "CCUTest.dll";
```
amirebrahimi: Temp/CCUTest.dll - no need to delete

Tests/Editor/CCUTest.cs
```diff
@@ -0,0 +1,86 @@
...
+				Path.Combine(Path.Combine(Path.Combine(extensionsPath, "Unity"), "GUISystem"), "UnityEngine.UI.dll"),
+				Path.Combine(Path.Combine(Path.Combine(Path.Combine(extensionsPath, "Unity"), "GUISystem"), "Editor"), "UnityEditor.UI.dll"),
+				Path.Combine(Path.Combine(Path.Combine(Path.Combine(extensionsPath, "Unity"), "EditorTestsRunner"), "Editor"), "nunit.framework.dll"),
+				Path.Combine(Path.Combine(Path.Combine(Path.Combine(extensionsPath, "Unity"), "EditorTestsRunner"), "Editor"), "UnityEditor.EditorTestsRunner.dll")
+			};
+
+			//GetAllFiles(EditorApplication.applicationContentsPath, references, "*.dll"); // Need to weed out unmanaged dlls
+			//GetAllFiles(dependencyPath, references, "*.dll"); // Error on loading ICsharpCode.NRefactory
+
+			var sources = new List<string>();
```
amirebrahimi: var sources = Directory.GetFiles(path, searchPattern, SearchOption.AllDirectories);

Tests/Editor/CCUTest.cs
```diff
@@ -0,0 +1,86 @@
...
+				Debug.LogError(string.Format("CCUTest: Could not delete temp files: {0}", e.Message));
+			}
+
+			foreach (var s in output)
+			{
+				Assert.IsFalse(s.Contains("error"), string.Join("\n", output));
+			}
+		}
+
+		static void GetAllFiles(string path, List<string> files, string searchPattern)
```
amirebrahimi: Remove

ucb_prebuild.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
amirebrahimi: Why this file here? Is this used?

Tests/Editor/TagManagerTest.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Linq;
+using NUnit.Framework;
+using UnityEditor.Experimental.EditorVR.Utilities;
+
+namespace UnityEditor.Experimental.EditorVR.Tests
+{
+	[InitializeOnLoad]
+	public class TagManagerTest
+	{
+		// TODO: It is possible to make use of tags and forget to add a RequiresTagAttribute. Find a test for this
```
amirebrahimi: Either remove or implement

Assets/.keep
```diff
@@ -0,0 +1 @@
+Workaround for CloudBuild automation
```
amirebrahimi: Why?

ucb_prebuild.sh
mtschoen-unity: This is here to get CloudBuild to play along. We need to make it think that this is a Unity project, so it moves all of the files into the empty Assets folder. Once we get the process working (we need them to add the EVR preview editor build) we will try to refine it so that we can keep the repo clean.

Assets/.keep
mtschoen-unity: See above. CB bails out if the repo doesn't look like a Unity project. We can probably remove these folders and the ucb_preload script for now since we still can't actually use UCB for testing.

Tests/Editor/CCUTest.cs
mtschoen-unity: This did not work as expected. We should discuss in our live review.

Tests/Editor/BuildPlayerTest.cs
mtschoen-unity: I tried this again and did not get prompted. I tried with the setting enabled before running tests, and disabled beforehand. I tried the tests individually as well as in a batch.

Assets/.keep
amirebrahimi: Yes, go ahead and remove. We could always have a separate project for CloudBuild purposes that uses a submodule for EditorVR.

PR #33, opened 2017-01-12
First pass at refactor of EditorVR.cs

amirebrahimi:
Splits EditorVR into separate partial class files as a start to have a more manageable class.


Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -0,0 +1,355 @@
...
+using UnityEngine.Experimental.EditorVR.Actions;
+using UnityEngine.Experimental.EditorVR.Menus;
+using UnityEngine.Experimental.EditorVR.Modules;
+using UnityEngine.Experimental.EditorVR.Tools;
+using UnityEngine.Experimental.EditorVR.Utilities;
+using UnityEngine.InputNew;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	internal partial class EditorVR : MonoBehaviour
```
mtschoen-unity: Maybe call this file EditorVR.ConnectInterfaces.cs?

Scripts/Core/EditorVR.Actions.cs
```diff
@@ -0,0 +1,93 @@
+#if !UNITY_EDITORVR
+#pragma warning disable 67, 414, 649
```
mtschoen-unity: Rather than disabling the warnings, why not wrap all of the class variables in #if UNITY_EDITORVR? I just did rough pass to check for issues and found none. The one caveat is also wrapping usages of EditorVR.kDefaultHideFlags in #if UNITY_EDITORVR. I'll commit it locally in case we want to go this route.

Scripts/Core/EditorVR.Workspaces.cs
```diff
@@ -0,0 +1,136 @@
...
+					tester = tester
+				};
+
+				m_IntersectionModule.AddTester(tester);
+			}, false);
+
+			UpdatePlayerHandleMaps();
+		}
+
+				private void OnWorkspaceDestroyed(IWorkspace workspace)
```
mtschoen-unity: tabs

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -1,3 +1,6 @@
+#if !UNITY_EDITORVR
+#pragma warning disable 67, 414, 649
```
mtschoen-unity: Rather than disabling these for the whole file, we should wrap preProcessRaycastSource in an #if UNITY_EDITORVR.

Or it might make sense for this class to be public again, since it's actually pretty useful as a stand-alone module.

Scripts/UI/PinnedToolButton.cs
```diff
@@ -1,4 +1,7 @@
-﻿using System;
+﻿#if !UNITY_EDITORVR
+#pragma warning disable 67, 414, 649
```
mtschoen-unity: Same here, with selected. In fact, it selected is _actually_ unused. It's just not reported as such because it's a public field on a public class. Anyway when I remove selected and the warning disable, I don't get any warnings from this class

PR #34, opened 2017-01-12
Improve menu(main+radial) hiding/showing when interacting with workspaces.

dunity:
Add support for better detection of menu position when interacting with workspaces.  Add support for contextual hiding of menus via dot checks, allowing for menus to be visible when the user doesn't "intend" to interact with workspaces (the device is not pointed towards a workspace front/top panel, or is beyond a certain panel's bounds).  Fix improperly named "for" loop variables in UpdateMenuVisibilityNearWorkspaces() in development.


amirebrahimi:
Addresses issue #42

Issue #44, opened 2017-01-18
Inspector does not update after object is modified anywhere else


mtschoen-unity:
For example, if I move an object in the 2D editor or with the manipulator, the values in the inspector do not update.

mtschoen-unity:
Completed PR #45

mtschoen-unity:
Completed PR #45

PR #45, opened 2017-01-18
Inspector responds to object changes

mtschoen-unity:
This doesn't require a design review. Inspector uses and undocumented Undo system callback to pick up on object modifications. This also means that our manipulator and other inspectors should register undo changes. I only covered the manipulator and the inspector in this change.


Workspaces/Inspector/InspectorWorkspace.cs
```diff
@@ -26,6 +27,8 @@ public class InspectorWorkspace : Workspace, ISelectionChanged
 
 	bool m_IsLocked;
 
+	bool m_BlockUndoPostProcess;
```
amirebrahimi: Not a fan of this approach. I'd say that we should likely subscribe and unsubscribe as needed, but I'm sure I'll understand more when we discuss this live.

Workspaces/Inspector/InspectorWorkspace.cs
```diff
@@ -62,6 +66,8 @@ public override void Setup()
 
 		if (Selection.activeGameObject)
 			OnSelectionChanged();
+
+		Undo.postprocessModifications += PostprocessModifications;
```
amirebrahimi: Is this undocumented or deprecated? Didn't see it in the latest docs.

Workspaces/Inspector/Scripts/ListItems/InspectorObjectFieldItem.cs
```diff
@@ -62,14 +56,37 @@ public void ClearButton()
 		SetObject(null);
 	}
 
+	bool TestAssignability(Object obj)
```
amirebrahimi: IsAssignable

Tools/TransformTool/TransformTool.cs
```diff
@@ -363,6 +363,8 @@ public void ProcessInput(ActionMapInput input, Action<InputControl> consumeContr
 #if UNITY_EDITOR
 			if (m_PivotRotation == PivotRotation.Local) // Manipulator does not rotate when in global mode
 				manipulatorTransform.rotation = Quaternion.Slerp(manipulatorTransform.rotation, m_TargetRotation, kLazyFollowRotate * deltaTime);
+
+			UnityEditor.Undo.RecordObjects(Selection.transforms, "EditorVR Manipulator");
```
amirebrahimi: import UnityEditor, so this doesn't have to be fully qualified.

Instead of "EditorVR Manipulator", I'd suggest using our current Unity naming "Move" to match with what happens with the Transform tool in scene view.

amirebrahimi: I went back and looked at our prototype code and we had [Undo.IncrementCurrentGroup](https://docs.unity3d.com/ScriptReference/Undo.IncrementCurrentGroup.html). I think this was necessary, so that each move would be handled separately instead of all being grouped together. Can you try out Undos an see if they all end up being grouped together?

Workspaces/Inspector/Scripts/ListItems/InspectorVectorItem.cs
```diff
@@ -75,6 +75,10 @@ public bool SetValue(string input, int index)
 	{
 		float value;
 		if (!float.TryParse(input, out value)) return false;
+
+		blockUndoPostProcess();
+		Undo.RecordObject(data.serializedObject.targetObject, "EditorVR Inspector");
```
amirebrahimi: Check naming here to match with standard Unity when you modify a property.

Tools/TransformTool/TransformTool.cs
mtschoen-unity: They certainly are. I wasn't sure how to fix that, thanks.

mtschoen-unity: Also re: "Move," I wasn't sure if we want to differentiate our actions from regular editor actions.

I was hoping to be able to read this group name when processing undo actions, but it doesn't seem to come through.

amirebrahimi:
Where are we with this?

mtschoen-unity:
We should be good to do a live review. I removed the BlockUndo fix and dimplemented what we talked about.

I found that adding/deleting components doesn't actually trigger that UndoPostProcess thing (even though they do actually register undo actions), so I ended up just letting the listview respond properly to a missing component. As far as responding to an added component, we're still in need of a solution. I think this is fine, since we don't currently let you add components in EVR, and when we do, we can trigger an update at that point.

UnityLabs:
Ready for final review

mtschoen-unity:
Closes #44

PR #50, opened 2017-01-24
Added getters for the VR proxies that represent the headset and controllers

justinmcmichael:
We have two items in here (the second one was merged after this pull requrest was created): 
1 - In order to represent VR editor users to each other in scene fusion, we needed access to the local editor's headset and handset locations. 
2 - Objects created with the primitive and duplicate tools are selected by default. This mimicks the behavior of the prefab instance creation, and provides compatibility with plugins like Scene Fusion.

### DIFF
--- a/Actions/Clone.cs
+++ b/Actions/Clone.cs
@@ -13,6 +13,8 @@ namespace UnityEngine.Experimental.EditorVR.Actions
 		public override void ExecuteAction()
 		{
 			var selection = Selection.gameObjects;
+			GameObject[] clones = new GameObject[selection.Length];
+			int index = 0;
 			var bounds = U.Object.GetBounds(selection);
 			foreach (var s in selection)
 			{
@@ -24,7 +26,9 @@ namespace UnityEngine.Experimental.EditorVR.Actions
 				cloneTransform.position = cameraTransform.TransformPoint(Vector3.forward * viewDirection.magnitude)
 					+ cloneTransform.position - bounds.center;
 				addToSpatialHash(clone);
+				clones[index++] = clone;
 			}
+			Selection.objects = clones;
 		}
 	}
 }
--- a/Actions/Clone.cs
+++ b/Actions/Clone.cs
@@ -13,8 +13,8 @@ namespace UnityEngine.Experimental.EditorVR.Actions
 		public override void ExecuteAction()
 		{
 			var selection = Selection.gameObjects;
-			GameObject[] clones = new GameObject[selection.Length];
-			int index = 0;
+			var clones = new GameObject[selection.Length];
+			var index = 0;
 			var bounds = U.Object.GetBounds(selection);
 			foreach (var s in selection)
 			{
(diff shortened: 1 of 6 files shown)


Actions/Clone.cs
```diff
@@ -13,6 +13,8 @@ public class Clone : BaseAction, IUsesSpatialHash
 		public override void ExecuteAction()
 		{
 			var selection = Selection.gameObjects;
+			GameObject[] clones = new GameObject[selection.Length];
```
amirebrahimi: var clones = new GameObject[selection.Length];

Actions/Clone.cs
```diff
@@ -13,6 +13,8 @@ public class Clone : BaseAction, IUsesSpatialHash
...
+			int index = 0;
```
amirebrahimi: var index = 0;

Actions/Paste.cs
```diff
@@ -37,6 +38,8 @@ public override void ExecuteAction()
 
 			if (buffer != null)
 			{
+				GameObject[] pastedGameObjects = new GameObject[buffer.Length];
```
amirebrahimi: var pastedGameObjects = ...

Actions/Paste.cs
```diff
@@ -37,6 +38,8 @@ public override void ExecuteAction()
...
+				int index = 0;
```
amirebrahimi: var index = 0;

Actions/Paste.cs
```diff
@@ -48,6 +51,11 @@ public override void ExecuteAction()
 						+ pastedTransform.position - bounds.center;
 					pasted.SetActive(true);
 					addToSpatialHash(pasted);
+					pastedGameObjects[index++] = pasted;
+				}
+				if (pastedGameObjects.Length > 0)
+				{
```
amirebrahimi: Remove braces for single line ifs

Actions/Paste.cs
```diff
@@ -48,6 +51,11 @@ public override void ExecuteAction()
 						+ pastedTransform.position - bounds.center;
 					pasted.SetActive(true);
 					addToSpatialHash(pasted);
+					pastedGameObjects[index++] = pasted;
+				}
+				if (pastedGameObjects.Length > 0)
```
amirebrahimi: new line

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -22,6 +22,10 @@ partial class EditorVR
 		DefaultProxyRay m_ProxyRayPrefab;
 
 		readonly Dictionary<Transform, DefaultProxyRay> m_DefaultRays = new Dictionary<Transform, DefaultProxyRay>();
+		public List<IProxy> proxies
```
amirebrahimi: EditorVR is an internal class, so we're not looking to expose it as a singleton. Would it be possible for you to collect all proxies via FindObjectsOfType?

Scripts/Core/EditorVR.cs
```diff
@@ -357,6 +357,10 @@ void AddPlayerModel()
 		}
 
 #if UNITY_EDITOR
+		public static EditorVR instance
```
amirebrahimi: See previous comment about `proxies`.

justinmcmichael:
I think we have a solution that is workable and doesn't require anything more than having objects be selected when created.

PR #55, opened 2017-02-10
EditorVR internal re-org (pass 2)

amirebrahimi:


Scripts/Core/EditorVR.Viewer.cs
```diff
@@ -0,0 +1,83 @@
+#if UNITY_EDITORVR
```
mtschoen-unity: I think we should call the file/class PlayArea, not Viewer. ViewerPivot sort of makes sense, but I think Play Area is more explicit about what you use it for.

Scripts/Input/OVRTouchInputToEvents.cs
```diff
@@ -1,164 +1,183 @@
...
-	public bool active { get; private set; }
-
-#if ENABLE_OVR_INPUT
-	public void Update()
-	{
-		// Manually update the Touch input
-		OVRInput.Update();
-
-		if ((OVRInput.GetActiveController() & OVRInput.Controller.Touch) == 0)
+		public bool active
```
mtschoen-unity: Can this go in a base class?

Scripts/Core/EditorVR.Viewer.cs
amirebrahimi: Rename references to ViewerPivot to CameraRig

Scripts/Core/EditorVR.DirectSelection.cs
```diff
@@ -1,221 +1,152 @@
...
-			cameraDiff.y = 0;
-			var rotationOffset = rotationDiff * cameraDiff - cameraDiff;
-
-			var endPosition = viewerPivot.position + (playerHead.position - mainCamera.position) + rotationOffset;
-			var endRotation = viewerPivot.rotation * rotationDiff;
-			var startTime = Time.realtimeSinceStartup;
-			var diffTime = 0f;
-
-			while (diffTime < kViewerPivotTransitionTime)
+			internal void OnObjectGrabbed(GameObject selection)
```
amirebrahimi: VS oddity, but technically would compile with static.

Scripts/Core/EditorVR.cs
```diff
@@ -1,7 +1,6 @@
 #if UNITY_EDITOR
 //#define ENABLE_MINIWORLD_RAY_SELECTION
```
amirebrahimi: Let's remove this.

Scripts/Core/EditorVR.cs
```diff
@@ -38,23 +37,83 @@ partial class EditorVR
...
+			public ITool currentTool;
+			public IMenu customMenu;
+			public PinnedToolButton previousToolButton;
+			public readonly Dictionary<IMenu, Menus.MenuHideFlags> menuHideFlags = new Dictionary<IMenu, Menus.MenuHideFlags>();
+			public readonly Dictionary<IMenu, float> menuSizes = new Dictionary<IMenu, float>();
+		}
+
+		class Nested
+		{
+			public static EditorVR evr { protected get; set; }
```
amirebrahimi: Constrained to EditorVR, so easy to update if we need that later.

Scripts/Modules/ObjectModule.cs
```diff
@@ -1,22 +1,30 @@
 ﻿using System;
```
amirebrahimi: Also rename PlaceObject

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -95,7 +95,7 @@ public override void Process()
 			// We don't process with all other input modules because we need fine-grained control to consume input
 		}
 
-		public void ProcessInput(ActionMapInput input, Action<InputControl> consumeControl)
+		public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
```
amirebrahimi: Moving towards a dictionary of IProcessInputs, so we'll keep this.

Issue #32, opened 2017-01-12
Tooltips


mtschoen-unity:
We need a way of displaying tooltips throughout the system. A tooltip is a piece of text that will appear when a user hovers over a button. Tooltips will automatically rotate to face the camera, and appear at some position offset relative to the button.

This feature will consist of a ToolTip component which will show/hide the tooltip, and a ToolTipAttribute for specifying tooltip data on dynamically generated UIs.

Because our UI is a mixture of hand-made prefabs and dynamically generated content (i.e. "View All" button on workspaces vs. RadialMenuSlots), there will be two way of defining a Tooltip. On hand-made prefabs, the author will simply add a ToolTip component and set the positional offset as a Vector3 field in the inspector. For dynamically-generated UI, developers will add the ToolTipAttribute with the text string to be displayed. The code which generates these UIs will be modified to check for the ToolTipAttribute and, if present, add a ToolTip component and specify the proper offset based on contect. In this way, the Radial Menu can decide where its ToolTips go based on the position of the button, which could change.

timoni:
- Tooltip angle should be based on initial HMD position, but not billboard.
- On the radial, tooltips should be positioned near the respective action.
- Tooltips should appear at the same time as the rest of the hover states.
- Tooltips should scale in.

mtschoen-unity:
Edit: Tooltips should show up aligned with their target transform, i.e. in the same plane as the Radial menu and aligned with the drag arrows in front of the workspace

timoni:
![tooltips](https://cloud.githubusercontent.com/assets/40673/22170543/c50a07ec-df33-11e6-8ba9-82ee5e360676.png)

PR #56, opened 2017-02-11
Tooltips

mtschoen-unity:
Adds tooltips to radial menu, workspaces, and pinned tool button

ITooltip specifies a string, a source transform, a target transform, and a TextAlignment property. Tooltips will cause errors if they do not specify a source or target transform.

Tooltip Module subscribes to rayEnter and rayExit events of Input Module. If a ray enters/exits a gameobject with a component that extends ITooltip, it shows/hides tooltip. Other systems can implement the IUsesTooltip interface, it will be given access to the show/hide methods on the Tooltip Module.

Tooltips can be added directly to prefabs with the Tooltip component. Existing UI components can implement ITooltip to provide tooltip information (string, transforms, alignment). A new TooltipAttribute provides the ability to decorate classes with a tooltip string, which is required in the case of Actions.

RadialMenuSlot has an interesting implementation because not only does it need to check for Actions with a Tooltip attribute, but also use an optional ITooltip reference for actions which can change their state like Switch Manipulator, etc.

Misc. fixes include a fix to materials on the Inspector lock button, as well as backported fixes related to world scaling from the improvements/schoen/locomotiontool branch.


amirebrahimi:
@mtschoen-unity, please resolve conflicts

Workspaces/MiniWorld/MiniWorldWorkspace.cs
```diff
@@ -122,6 +124,10 @@ public override void Setup()
 			connectInterfaces(mb);
 		}
 
+		var zoomTooltip = sliderObject.GetComponentInChildren<Tooltip>();
```
amirebrahimi: Shouldn't this be ITooltip?

Workspaces/Project/ProjectWorkspace.cs
```diff
@@ -98,6 +99,10 @@ public override void Setup()
 			connectInterfaces(mb);
 		}
 
+		var zoomTooltip = sliderObject.GetComponentInChildren<Tooltip>();
```
amirebrahimi: ITooltip

Actions/Copy.cs
```diff
@@ -1,9 +1,9 @@
 ﻿using UnityEditor;
-using UnityEngine.Experimental.EditorVR.Utilities;
 
 namespace UnityEngine.Experimental.EditorVR.Actions
 {
 	[ActionMenuItem("Copy", ActionMenuItemAttribute.kDefaultActionSectionName, 5)]
+	[Tooltip("Copy")]
```
amirebrahimi: Since actions seem to be the only one that make use of this attribute, I think we can get rid of the attribute and fold the tooltip into the ActionMenuItem attribute as an additional parameter and/or simply use the text.

Editor/TooltipsEditor.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System;
```
amirebrahimi: I haven't tried this out yet directly, but am wondering what the custom editor delivers?

Scripts/Interfaces/ITooltip.cs
```diff
@@ -0,0 +1,28 @@
...
+		/// </summary>
+		string tooltipText { get; }
+
+		/// <summary>
+		/// The transform relative to which the tooltip will display
+		/// </summary>
+		Transform tooltipTarget { get; }
+
+		/// <summary>
+		/// The transform relative to which the tooltip will display
```
amirebrahimi: tooltipTarget and tooltipSource share the same summaries; what is the difference between these two?

Scripts/Interfaces/IUsesToolTip.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace UnityEngine.Experimental.EditorVR
+{
+	/// <summary>
+	/// Provides access to the ability to show or hide a Tooltip
+	/// </summary>
+	public interface IUsesTooltip
```
amirebrahimi: Possibly ISetTooltipVisibility

amirebrahimi: And it looks like it's necessary for anything that wants to show tooltips to use this method? No tooltips would show automatically?

Scripts/Modules/TooltipModule/Tooltip.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace UnityEngine.Experimental.EditorVR.UI
+{
+	public class Tooltip : MonoBehaviour, ITooltip
```
amirebrahimi: Mark this internal and any other new classes you've created in this PR.

Scripts/Interfaces/IUsesToolTip.cs
mtschoen-unity: Not exactly. We add show/hide to rayEnter/Exit on the input module so that any UI with an ITooltip in the hierarchy will show its tooltip.

Editor/TooltipsEditor.cs
mtschoen-unity: It is a centralized place to edit serialized tooltip strings and show the user where tooltips exist within the codebase. It's a potential first step toward being able to localize Tooltips, and generally helps us keep track of them.

Actions/Copy.cs
mtschoen-unity: I was also considering having Action implement IToolTip and just hardcoding the text as a property. We probably want to still allow for the item name and tooltip text to be different, so I'm also amenable to an optional string parameter.

Workspaces/Project/ProjectWorkspace.cs
mtschoen-unity: The problem is that ITooltip doesn't provide a setter on tooltipText. We can discuss whether this needs to change.

mtschoen-unity:
I've made the suggested changes and I'm ready to do a live review.

Scripts/UI/GradientButton.cs
```diff
@@ -147,7 +146,7 @@ public bool visible
 		Graphic[] m_HighlightItems;
 
 		[Header("Animated Reveal Settings")]
-		[Tooltip("Default value is 0.25")]
+		[UnityEngine.Tooltip("Default value is 0.25")]
```
amirebrahimi: This can go back to what it was.

Scripts/Core/EditorVR.cs
```diff
@@ -150,6 +151,11 @@ void Awake()
 			m_InputModule.dragStarted += m_DragAndDropModule.OnDragStarted;
 			m_InputModule.dragEnded += m_DragAndDropModule.OnDragEnded;
 
+			m_TooltipModule = U.Object.AddComponent<TooltipModule>(gameObject);
```
amirebrahimi: AddModule<TooltipModule>

Scripts/Interfaces/IUsesToolTip.cs
amirebrahimi: On second thought, I think ISetTooltipVisibility is more descriptive.

mtschoen-unity:
Closes #32

PR #58, opened 2017-02-16
first pass at hacking in the context stack

leith-bartrich:
-Moved some code form EditorVR to VRView and added the context stack.  This decouples VRView from EditorVR.

-Added IEditorContext and applied it to EditorVR.

-Added the optional concept of Configuration to Editing Contexts.

This is for discussion purposes.  I have tested it to the extent that I've confirmed that the existing EditorVR still runs with this rewiring.  The integration tests you mentioned on the hangout earlier are a thing I don't have access to.  So I can't confirm if I've created a regression.

This means that Sub-contexts and alternative contexts are not currently tested.  Though you can see how they're supposed to work.

I've left the hotkeys alone and I have not authored a new "root context."  Rather, I've done the minimum wiring to allow such a thing to be built.

Next steps would be to pull an abstract base-class or three from EditorVR from which to inherit and create an alternative Editing Context or two.

Thoughts?

This isn't much code.  I don't want to put too much more effort into it before we've discussed.  We might decide to ditch it and re-write.

### DIFF
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -24,9 +24,8 @@ namespace UnityEditor.Experimental.EditorVR
 	[InitializeOnLoad]
 #if UNITY_EDITORVR
 	[RequiresTag(kVRPlayerTag)]
-	partial class EditorVR
+	partial class EditorVR : IEditingContext
 	{
-		public const HideFlags kDefaultHideFlags = HideFlags.DontSave;
 		const string kVRPlayerTag = "VRPlayer";
 
 		[SerializeField]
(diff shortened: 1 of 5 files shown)


mtschoen-unity:
Interesting choices. It looks like with Subvert/Revive that you are already intending to swap contexts at "runtime" (e.g. without closing VRView).

I'm also curious about why you kept the inputmanager around as a part of VRView. Why not also make that into a re-usable module that we can optionally override in the Configure phase?

I'll wait for the rest of the team to take a gander at this. I'm not sure where this will fit on our priority queue, but thanks for giving us a concrete idea of what you're talking about! 👍

leith-bartrich:
I was questioning the InputManager's home as I did it.  I probably need a better sense of the new Unity Input system's design before I could make a better decision.  But I could easily see it left to the context as you suggest.  Maybe rolled back into an abstract baseclass pulled out of EditorVR.  Or injected via the Config if that makes sense for the given Context.

Yes, as you surmise I intend to swap contexts at "run/edit-time."

The two contexts I'd try and successfully author first, as test cases, would likely be:

-MyShellContext
-MySpecifcEditingContext

Where the MyShellContext is a top level launcher for getting into "Unity Scene Editor" and "MySpecificEdtingContext." As well as any other enumerated Contexts that mark themselves as arbitrarily launch-able by a ShellContext.

And MySpecificEditingContext would likely be some kind of specific content editing context.  Such as maybe a context for editing terrain.  Or lighting.  In an alternative and focused way.

amirebrahimi:
Hi @leith-bartrich, thank you for providing this. I'm switched this PR to pull into a custom branch, so that I can update it to latest development and continue to make some changes. I'm going to add you back on the PR for it.

I wasn't considering that editing contexts would be hot-swappable, but will keep that as a goal.

PR #59, opened 2017-02-16
Locomotion improvements

### LINKED ISSUE #9: Allow scaling of the world from viewer's perspective
Currently it's only possible to have 1:1 scale from the viewer's perspective. We have the MiniWorld to allow scaling the view of the scene for manipulation purposes. This feature would allow scaling the view of the world around you.

### LINKED ISSUE #17: Improve locomotion to allow for change in Z
Feedback from user:
Most of the movement system seemed to focus on X/Y movement which made it hard as I like to use a lot Z Movement when moving around. The teleportation system worked however teleportation is not very precise/immersive. I found it very useful to cross long distances but again it was only on the X/Y Plane. If the Teleporter was directional rather than a typical arc teleporter I think I would find it more useful. That way I could teleport up and Down by pointing in that direction.
There is a movement system on the asset store we use quite a bit for navigating around in 3D space. The Rubber band movement works like a 3D Joystick allowing control your velocity quite precisely. They also have a flight system that uses the rotation of the controller to move around. I would take a look at it to see what you think.

mtschoen-unity:
These changes primarily add two-handed world/player scaling. This happens in BlinkLocomotoionTool.cs.  The basic principal is that you when you hold both grip button/triggers, we record the starting position of the controllers, and then translate/rotate/scale the camera rig (formerly viewer pivot) based on the offset.

Pulling your hands apart reduces the scale of the camera rig, making the world appear to be larger, while pulling them together scales it up. Translation is calculated based on the change in position of the midpoint  between the controllers. That way if you move both hands, the world moves 1:1 with that motion, while moving only one hand moves the world proportionally. The rotation is also applied around the midpoint, and constrained to yaw-only, which ensures that any object that starts out between the controllers stays there. In our example scene, if you put each controller inside of a tree, the trees will stay locked to the controllers while you rotate and scale, which is very satisfying!

Scaling is done by simply multiplying the camera rig's current local scale by the ratio of the current distance between controllers to what it was when it started. Things start to get weird when you can translate and rotate while you're scaling. You have to apply the scale factor to the original offsets in the inverse while taking into account current world-scale offsets. The important thing to avoid shakiness/unstable results is to scale the offsets by the current scale, not the scale factor.

I introduced ILinkedTool which gives tools access to all other instances of the same tool, as well as a bool to designate the primary (first created) tool. IHandedTool is used to reliably check which hand a given tool is bound to, in order to provide a reliable user experience. It is used for the changes to Joystick Locomotion.

Visuals are a self-contained prefab/MonoBehaviour that take the left/right hands and cameraRig to display a dotted line between the hands and a sliding scale representing your current scale.

I also changed the way that joystick locomotion works: One hand moves up and down while the other moves forward and backward. Also, instead of using a quadratic filter to quickly ramp up movement/rotation speed, I introduced a threshold above which a super-fast value is used. Below the threshold, values scale linearly, and above it, you move/rotate at a constant speed. In this way, you can cut out the middle-range of speeds which causes nausea. Low speeds are almost useless, though, so we might just want to use a single speed just fast enough to avoid nausea.

Other fixes are for systems which behave differently when the camera rig has been scaled. There might be a few other edge cases which need to be addressed. I decided to scale the clipping planes of the camera because the controllers ended up clipping at low scales. This might cause issues if the initial values change.

Resolves #9 
Resolves #17


Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -21,6 +22,8 @@ class Interfaces : Nested
 
 			readonly HashSet<object> m_ConnectedInterfaces = new HashSet<object>();
 
+			readonly Dictionary<Type, List<ILinkedTool>> m_LinkedTools = new Dictionary<Type, List<ILinkedTool>>();
```
amirebrahimi: This should probably go in EditorVR.Tools.cs now.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -40,13 +43,17 @@ internal void ConnectInterfaces(object obj, InputDevice device)
...
 					rayOrigin = deviceData.rayOrigin;
+					node = deviceData.node;
+				}
 
-				ConnectInterfaces(obj, rayOrigin);
+				ConnectInterfaces(obj, rayOrigin, node);
 			}
 
-			internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
+			internal void ConnectInterfaces(object obj, Transform rayOrigin = null, Node node = Node.LeftHand)
```
amirebrahimi: I would rather node be nullable.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -79,6 +86,10 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
 					if (ray != null)
 						ray.rayOrigin = rayOrigin;
 
+					var handedRay = obj as IUsesHandedRayOrigin;
```
amirebrahimi: Let's separate these two interfaces, since we already have IUsesRayOrigin. Might as well have IUsesNode.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -282,6 +293,27 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
...
+				var linkedTool = obj as ILinkedTool;
+				if (linkedTool != null)
+				{
+					List<ILinkedTool> otherTools;
+					var type = obj.GetType();
+					if (m_LinkedTools.TryGetValue(type, out otherTools))
+					{
+						foreach (var otherTool in otherTools)
+						{
+							otherTool.otherTools.Add(linkedTool);
```
amirebrahimi: We should simply maintain a list of all tools per linked tool instead of trying to exclude the single tool and maintain all the lists separately.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -282,6 +293,27 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
...
+						}
+
+						linkedTool.otherTools.AddRange(otherTools);
+					}
+					else
+					{
+						linkedTool.primary = true;
```
amirebrahimi: I'm not clear on what makes a tool primary. My preference would be to remove this distinction.

Scripts/Interfaces/ILinkedTool.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Provides access to other tools of the same type
+/// </summary>
+public interface ILinkedTool
+{
+	/// <summary>
+	/// List of other tools of the same type (not including this one)
+	/// </summary>
+	List<ILinkedTool> otherTools { get; }
```
amirebrahimi: linkedTools or toolInstances

Scripts/Interfaces/IUsesHandedRayOrigin.cs
```diff
@@ -0,0 +1,15 @@
+﻿using UnityEngine.Experimental.EditorVR.Tools;
+
+namespace UnityEngine.Experimental.EditorVR
+{
+	/// <summary>
+	/// Adds Node information to IUsesRayOrigin to determine which hand the tool is attached to
+	/// </summary>
+	public interface IUsesHandedRayOrigin : IUsesRayOrigin
```
amirebrahimi: I suppose we could still have IUsesHandedRayOrigin, but it should combine IUsesRayOrigin and IUsesNode.

Tools/BlinkLocomotionTool/BlinkLocomotionTool.cs
```diff
@@ -29,28 +47,60 @@ private enum State
...
+	bool m_AllowScaling = true;
+	bool m_Scaling;
+	float m_StartScale;
+	float m_StartDistance;
+	Vector3 m_StartPosition;
+	Vector3 m_StartMidPoint;
+	Vector3 m_StartDirection;
+	float m_StartYaw;
+
+	InputControl m_Grip;
```
amirebrahimi: Why is this cached?

Tools/BlinkLocomotionTool/Scripts/WorldScaleVisuals.cs
```diff
@@ -0,0 +1,81 @@
+﻿using UnityEngine;
+using UnityEngine.Experimental.EditorVR.Tools;
+using UnityEngine.Experimental.EditorVR.Utilities;
+using UnityEngine.UI;
+
+public class WorldScaleVisuals : MonoBehaviour, IUsesCameraRig
```
amirebrahimi: Mark this internal + any other new classes you've created
Add a proper namespace for this to live in.

timoni:
Icon pack for scaling. https://cl.ly/jCgV or 
atom
burrito
person
sea monster
building
moon

mtschoen-unity:
This closes #17 and #9

Scripts/Core/EditorVR.Interfaces.cs
mtschoen-unity: The primary tool is just the first one which is created. This is needed in order to do "cross tool" procedures only once per frame, rather than once per frame per tool. We could rename it to "first" or "master" if that helps.

Scripts/Interfaces/ILinkedTool.cs
mtschoen-unity: Actually, I'm starting to think that this goes for more than just tools. We could use this in MiniWorldWorkspace to move some of the logic out of EditorVR.MiniWorlds

Tools/BlinkLocomotionTool/Scripts/BlinkVisuals.cs
```diff
@@ -78,6 +79,8 @@ private enum State
 	public bool validTarget { get; private set; }
 	public bool showValidTargetIndicator { private get; set; }
 
+	public Transform cameraRig { get; set; }
```
amirebrahimi: Have BlinkTransformTool set a transform if needed or optionally, just provide the scale.

Tools/BlinkLocomotionTool/Scripts/WorldScaleVisuals.cs
```diff
@@ -0,0 +1,83 @@
...
+	Sprite[] m_Icons;
+
+	[SerializeField]
+	GameObject m_IconPrefab;
+
+	float m_LineWidth;
+
+	public Transform leftHand { private get; set; }
+	public Transform rightHand { private get; set; }
+	public Transform cameraRig { private get; set; }
```
amirebrahimi: Same here. Just have BlinkTransformTool set this.

Scripts/Core/EditorVR.DirectSelection.cs
```diff
@@ -37,9 +37,9 @@ internal float GetPointerLength(Transform rayOrigin)
 					{
 						var miniWorld = ray.miniWorld;
 
-						// As the miniworld gets smaller, the ray length grows, hence localScale.Inverse().
+						// As the miniworld gets smaller, the ray length grows, hence lossyScale.Inverse().
 						// Assume that both transforms have uniform scale, so we just need .x
-						length *= miniWorld.referenceTransform.TransformVector(miniWorld.miniWorldTransform.localScale.Inverse()).x;
+						length *= miniWorld.referenceTransform.TransformVector(miniWorld.miniWorldTransform.lossyScale.Inverse()).x;
```
amirebrahimi: To discuss

mtschoen-unity: Yep. There are a few more places where I needed to replace localScale with lossyScale, in case you want to check those out ahead of time.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -290,6 +290,10 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
 				if (moveCameraRig != null)
 					moveCameraRig.moveCameraRig = Viewer.MoveCameraRig;
 
+				var usesViewerScale = obj as IUsesViewerScale;
```
amirebrahimi: IUsesWorldScale

Scripts/Core/EditorVR.Viewer.cs
```diff
@@ -102,6 +102,11 @@ internal static void MoveCameraRig(Vector3 position, Vector3? viewdirection)
 			{
 				evr.StartCoroutine(UpdateCameraRig(position, viewdirection));
 			}
+
+			internal static float GetViewerScale()
```
amirebrahimi: GetWorldScale

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -386,20 +386,23 @@ private void Select(GameObject go)
 
 		private GameObject GetRayIntersection(RaycastSource source)
 		{
-			GameObject hit = null;
 			// Move camera to position and rotation for the ray origin
 			m_EventCamera.transform.position = source.rayOrigin.position;
 			m_EventCamera.transform.rotation = source.rayOrigin.rotation;
 
-			RayEventData eventData = source.eventData;
+			var camera = U.Camera.GetMainCamera();
```
amirebrahimi: Add a comment OR is this even necessary? Do we need to change far clip or was it really just about adjusting the near clip to a reasonable value?

Tools/BlinkLocomotionTool/Scripts/BlinkVisuals.cs
```diff
@@ -78,7 +78,7 @@ private enum State
 	public bool validTarget { get; private set; }
 	public bool showValidTargetIndicator { private get; set; }
 
-	public Transform cameraRig { get; set; }
+	public float viewerScale { private get; set; }
```
amirebrahimi: worldScale

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
mtschoen-unity: We can probably get away with just adjusting near clip, but I was worried about losing depth precision when increasing the ratio of far/near.

Will add a comment about the fact that we are doing this for world scale purposes

mtschoen-unity:
So yeah, viewer scale vs world scale. They are inverse of each other and sometimes I use viewer and sometimes I use the inverse, depending on the situation. The actual value you are getting _is_ the viewer scale. I could instead change the places where I call it "inverseViewerScale" to "worldScale"

PR #72, opened 2017-02-23
Staging review changes

### LINKED ISSUE #6: Move all public api interfaces to UnityEditor.Experimental.EditorVR
Previously, we had these split across Tools, Workspaces, Actions, etc., but from our preferred def usage we'd like to have someone have one simple `using UnityEngine.Experimental.EditorVR`.

amirebrahimi:
This is a PR to address review notes and will be merged back into staging, upon which staging would hopefully be approved.

What's in this PR:
- Mark classes internal/sealed
- Move all global classes into namespaces
- Change delegate fields to have no initialization
- Remove 'k' prefix from public fields constants
- Fix static fields that didn't use `s_` convention
- Fix constant fields to use `k_` convention 
- Rename U.* utility classes to match *Utils convention
- Wrap all classes in UNITY_EDITOR to protect against being included in builds

Resolves #6


Scripts/Input/OVRTouchInputToEvents.cs
```diff
@@ -1,164 +1,167 @@
...
+using System;
+using UnityEditor.Experimental.EditorVR;
 
 [assembly: OptionalDependency("OVRInput", "ENABLE_OVR_INPUT")]
 
-/// <summary>
-/// Sends events to the input system based on native Oculus SDK calls
-/// </summary>
-public class OVRTouchInputToEvents : MonoBehaviour
+namespace UnityEngine.InputNew
```
adrianoc-unity3d: s this namespace correct? Shouldn't it be _UnityEngine.Experimental._**something**_.Input_ ?

Scripts/Input/OVRTouchInputToEvents.cs
```diff
@@ -1,164 +1,167 @@
...
 {
+	/// <summary>
+	/// Sends events to the input system based on native Oculus SDK calls
+	/// </summary>
+	public class OVRTouchInputToEvents : MonoBehaviour
```
adrianoc-unity3d: was this class intentionally left as **public** ?

Scripts/Interfaces/IUsesSpatialHash.cs
```diff
@@ -1,18 +1,23 @@
...
 
-/// <summary>
-/// Make use of the spatial hash
-/// </summary>
-public interface IUsesSpatialHash
+namespace UnityEditor.Experimental.EditorVR
 {
 	/// <summary>
-	/// Add all renderers of a GameObject (and its children) to the spatial hash for queries, direct selection, etc.
+	/// Make use of the spatial hash
```
adrianoc-unity3d: not directly related to this PR but I think it worth mentioning: 

can we come up with better names for interfaces named like  **IUses**_Whatever_ ?

Maybe **ISupport**_Whatever_ ?

Scripts/Input/OVRTouchInputToEvents.cs
amirebrahimi: I left it in the default namespace that the input-prototype uses. Originally, this was in the input-prototype repo and we pulled it out at one point. All of this code will get replaced soon with the new input system, including the whole UnityEngine.InputNew namespace.

Scripts/Input/OVRTouchInputToEvents.cs
amirebrahimi: Yes, for the same reasoning as above. I could move this back into the input-prototype repository at some point, too, but it isn't a high priority.

Scripts/Modules/ObjectPlacementModule.cs
```diff
@@ -11,8 +11,8 @@ internal class ObjectPlacementModule : MonoBehaviour, IUsesSpatialHash
 		const float kInstantiateFOVDifference = -5f;
 		const float kGrowDuration = 0.5f;
 
-		public Action<GameObject> addToSpatialHash { get; set; }
-		public Action<GameObject> removeFromSpatialHash { get; set; }
+		public Action<GameObject> addToSpatialHash { private get; set; }
```
amirebrahimi: That would require explicit interface declaration and access, no? e.g. IUsesSpatialHash.addToSpatialHash { ... }

Scripts/Input/OVRTouchInputToEvents.cs
adrianoc-unity3d: ok, I don't think this will be necessary

PR #74, opened 2017-02-27
Fix Vive controller model not showing up

mtschoen-unity:
Turns out deactivating the controller object canceled the SetModelAsync coroutine, causing the model load to fail


Scripts/Input/ViveInputToEvents.cs
```diff
@@ -39,7 +39,7 @@ static ViveInputToEvents()
 
 		public void Update()
 		{
-			active = false;
+			var shouldAactivate = false;
```
amirebrahimi: Typo here, but I also think that isActive is more appropriate.

PR #75, opened 2017-03-03
UI Shader Structure improvements

### LINKED ISSUE #77: Fix occlusion issues with workspaces, menus, and manipulators
Re-worked shaders / materials to correct these issues and are now more simplified.

dunity:
In preparation for tool switch/flow improvements, and to address the current issues in the development branch with menus/workspaces/manipulators drawing above/through each other at times, the render-queues and shader structures were refactored to have better respect for depth, with less emphasis on requiring stencils in some cases.

- WorkspaceBase root-canvas sorting order set to 0
- Stencil block was removed from from ListClip, and ListClipText shaders
- New WorkspaceUI & WorkspaceUIMask materials added; set on graphic components that need custom queue order
- The FilterUI position bug was also fixed.  It is no longer inside workspaces
- Fixed FocusUI prefab's text tracking & size
- Blending & alpha support added to VerticalBorderGradient shader

Resolves #77


Prefabs/UI/Materials/TooltipBackground.mat
```diff
@@ -7,18 +7,18 @@ Material:
...
+  m_CustomRenderQueue: 9002
   stringTagMap: {}
   m_SavedProperties:
     serializedVersion: 2
     m_TexEnvs:
     - first:
         name: _BumpMap
       second:
-        m_Texture: {fileID: 2800000, guid: 06dc8da358c6ebb4dad67b286b513f95, type: 3}
+        m_Texture: {fileID: 0}
```
amirebrahimi: Is this deliberate? Just noticing that there was a texture assigned before.

Workspaces/Common/Scripts/WorkspaceButton.cs
```diff
@@ -334,7 +334,7 @@ IEnumerator BeginHighlight()
 			var highlightedLocalScale = new Vector3(transform.localScale.x, transform.localScale.y, m_VisibleLocalZScale * 2);
 			while (transitionAmount < kTargetTransitionAmount)
 			{
-				transitionAmount += Time.unscaledDeltaTime * 3;
+				transitionAmount += Time.unscaledDeltaTime * 5;
```
amirebrahimi: Needs comment or magic number needs to be in a variable

Prefabs/UI/Materials/TooltipBackground.mat
dunity: It is deliberate.  The texture had to be reassigned to the albedo slot after the shader swap to the standard shader.

PR #76, opened 2017-03-07
Merge back 0.0.1 scripting review changes to development

amirebrahimi:
These changes were primarily made in the staging branch prior to 0.0.1 release. Now, they need to be merged back to development. This is required before 0.0.2 can be released.


Scripts/Modules/SpatialHashModule.cs
```diff
@@ -19,11 +21,11 @@ void Awake()
 
 		internal void Setup()
 		{
-			SetupObjects();
+			SetObjectUtilss();
 			StartCoroutine(UpdateDynamicObjects());
 		}
 
-		void SetupObjects()
+		void SetObjectUtilss()
```
mtschoen-unity: typo

Scripts/Modules/Highlight/HighlightModule.cs
```diff
@@ -24,7 +24,7 @@ void LateUpdate()
 				foreach (var m in go.GetComponentsInChildren<MeshFilter>())
 				{
 					var highlightTransform = m.transform;
-					Matrix4x4 highlightScaleIncreaseMatrix = Matrix4x4.TRS(highlightTransform.position, highlightTransform.rotation, highlightTransform.lossyScale + kHighlightScaleIncrease);
+					Matrix4x4 highlightScaleIncreaseMatrix = Matrix4x4.TRS(highlightTransform.position, highlightTransform.rotation, highlightTransform.lossyScale + k_HighlightScaleIncrease);
```
mtschoen-unity: I just realized this also wants a * Viewer.GetViewerScale(). We can add this later.

Scripts/Modules/SpatialHashModule.cs
amirebrahimi: Good catch. Search and replace

PR #80, opened 2017-03-08
Fix ViveProxy initialization when missing SteamVR

dunity:
Fix null m_InputToEvents exception (& material exceptions) when starting EditorVR without SteamVR installed

### DIFF
--- a/Scripts/Proxies/ViveProxy.cs
+++ b/Scripts/Proxies/ViveProxy.cs
@@ -10,13 +10,18 @@ namespace UnityEditor.Experimental.EditorVR.Proxies
 #if ENABLE_STEAMVR_INPUT
 		SteamVR_RenderModel m_RightModel;
 		SteamVR_RenderModel m_LeftModel;
+#endif
 
 		public override void Awake()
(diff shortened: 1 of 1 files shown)


Scripts/Proxies/ViveProxy.cs
```diff
@@ -10,13 +10,18 @@ sealed class ViveProxy : TwoHandedProxyBase
 #if ENABLE_STEAMVR_INPUT
 		SteamVR_RenderModel m_RightModel;
 		SteamVR_RenderModel m_LeftModel;
+#endif
 
 		public override void Awake()
 		{
 			base.Awake();
 			m_InputToEvents = ObjectUtils.AddComponent<ViveInputToEvents>(gameObject);
+#if !ENABLE_STEAMVR_INPUT
```
amirebrahimi: Can you test if this still works without this change and remove it if it does?

dunity: It's required to prevent a material related exception thrown in TwoHandedProxyBases' Start function

PR #81, opened 2017-03-08
Remove old VRLineRenderer library folder and replace it with a submodule

amirebrahimi:


.gitmodules
```diff
@@ -4,4 +4,7 @@ url=https://github.com/Unity-Technologies/input-prototype.git
 	branch = editorvr/main
 [submodule "libs/UnityOctree"]
 	path = libs/UnityOctree
-url=https://github.com/Unity-Technologies/UnityOctree.git
\ No newline at end of file
+url=https://github.com/Unity-Technologies/UnityOctree.git
```
mtschoen-unity: Not sure what happened here, but can we add a tab to this line?

PR #88, opened 2017-03-10
Hide MiniWorldReference and MiniWorldCamera from hierarchy

mtschoen-unity:
It looks like this issue has always been around. I thought about adding an IUsesHideFlags and actually supplying the proper hideflags value, but this would mean switching to a Start() instead of OnEnable since it would need the interface to be hooked up. I was also thinking of using static members and setting them when EVR sets itself up.

This seemed like the cleanest fix.

### DIFF
--- a/Workspaces/MiniWorld/Scripts/MiniWorld.cs
+++ b/Workspaces/MiniWorld/Scripts/MiniWorld.cs
@@ -84,7 +84,7 @@ namespace UnityEditor.Experimental.EditorVR.Workspaces
 		private void OnEnable()
 		{
 			if (!referenceTransform)
-				referenceTransform = new GameObject("MiniWorldReference") { hideFlags = HideFlags.DontSave }.transform;
+				referenceTransform = new GameObject("MiniWorldReference") { hideFlags = HideFlags.HideAndDontSave }.transform;
 
 			m_MiniWorldRenderer = ObjectUtils.AddComponent<MiniWorldRenderer>(CameraUtils.GetMainCamera().gameObject);
 			m_MiniWorldRenderer.miniWorld = this;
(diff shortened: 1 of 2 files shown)


Workspaces/MiniWorld/Scripts/MiniWorld.cs
```diff
@@ -96,7 +96,7 @@ private void OnEnable()
 
 		private void OnDisable()
 		{
-			if ((referenceTransform.hideFlags & HideFlags.DontSave) != 0)
+			if ((referenceTransform.hideFlags & HideFlags.HideAndDontSave) != 0)
```
amirebrahimi: Why this?

Workspaces/MiniWorld/Scripts/MiniWorld.cs
```diff
@@ -84,7 +84,7 @@ public List<Renderer> ignoreList
 		private void OnEnable()
 		{
 			if (!referenceTransform)
-				referenceTransform = new GameObject("MiniWorldReference") { hideFlags = HideFlags.DontSave }.transform;
+				referenceTransform = new GameObject("MiniWorldReference") { hideFlags = HideFlags.HideAndDontSave }.transform;
```
amirebrahimi: Use ObjectUtils.Instantiate here and it will get the hideFlags.

Workspaces/MiniWorld/Scripts/MiniWorldRenderer.cs
```diff
@@ -53,7 +53,7 @@ void OnEnable()
 		{
 			var go = new GameObject("MiniWorldCamera", typeof(Camera));
```
amirebrahimi: Should use ObjectUtils.Instantiate. I think at the time I was thinking we'd have this work in play mode, but that's a much bigger task to make everything work now.

Workspaces/MiniWorld/Scripts/MiniWorld.cs
mtschoen-unity: That's a good question, actually. I changed it so that it would successfully, destroy, but the hideflags check has always been there, and I think you might be the one who originally wrote it.

mtschoen-unity: I traced this back to the original version of the file. Since I can't think of a reason for it, I replaced it with just if (referenceTransform)

PR #94, opened 2017-03-16
Hierarchy drag and drop

### LINKED ISSUE #46: Drag and drop to re-arrange hierarchy
Like the 2D Hierarchy window, we should be able to drag and drop objects to change their order and parenting.

### LINKED ISSUE #92: Improve list views
The most noticeable change is that scrolling momentum works again. Some workspaces also had a redundant background panel causing a double-highlight. There were also cases where the list would snap back to the wrong position or get "stuck" when scrolling with the thumbstick.

Behind the scenes, much has changed to unify and re-use more code.

### LINKED ISSUE #93: Hierarchy filtering
We need the ability to filter the hierarchy view by component type

mtschoen-unity:
Implementing this feature required a major refactor to the List View classes, mostly because of the need to animate transitions between states.

This PR closes #46, closes #92, and closes #93

As for the ListView changes, one of the more major changes is renaming of all of the template types to match conventions (i.e. TIndex), as well as passing the Item type template through NestedListViewController which removed the need for casting in a few places. Speaking of TIndex, a generic index field was added to ListViewData which formalizes the concept of a unique index for mapping purposes. This allows me to move the map of visuals objects to the base ListViewController class and use it throughout all levels of the system. Another big change is to use a simple "visible length" for determining how many items tor draw, rather than keeping an integral data offset. I realized that for Inspector lists, which don't have uniform row sizes, this system already had the potential to display too few or too many rows. It also gets rid of the ugly if (scrollOffset < 0) m_DataOffset-- hack.

I refactored FilterUI to remove the static reference to the FilterUI class by adding an ExtensionMethods class to IFilterUI. We also separate ProjectData IFilterUIs vs HierarchyData IFilterUIs since they use a different list of types.

At @timoni's request, I re-worked how DraggableListItem works to simply use a distance threshold, rather than a double-click to detect a grab. InspectorListItem extends this functionality to differentiate between a horizontal (local space) drag vs a vertical drag (more specifically "anything but 85% horizontal"). This works much better across the board for Hierarchy drag/drop and Inspector field grabbing.

After the first design review, @timoni asked for transition states to indicate which rows are going where. This introduced the need for the "settling" feature, which I eventually decided to build into ListViews generally. The reason for this is that transitions and interpolation were part of the original goal of the framework, and we actually need some form of "settle" state for all list views. Eventually the folder list will allow dragging and dropping, and the Inspector already makes use of it for expanding/collapsing rows. Basically you call StartSettling after kicking off some sort of major change to the list, let the update method interpolate objects around, and as soon as all objects reach their destination, the settling state is unset, and an optional completion callback fires.

DragAndDropModule got a minor change. Because we change the state of whether you can drop the row after the initial drag, we can't store the drop object in OnDragStart, but instead have to check the IDroppable every drag frame. Thus, we store a list of IDroppables, instead of a list of objects.

I added a PointerClick events to MultipleRayInputModule and BaseHandle. I also refactored some of the code as I made modifications for caching/style purposes. I switched the order of ExecuteEvents and the module-level event call on end drag. The reason for this is that if the events fire first, the "release" event tells the object that it should not expose a drop object before the the DragAndDropModule calls OnDrop.

I'm not sure if this is new, but field dragging was messed up. I simplified the code and fixed some situations where undo modifications would interfere while you were dragging a field.

Some misc fixes not worth creating issues for:
* Inspector lock icon started in locked state
* Inspector text was not visible for all row types, or when fields were grabbed
* Added CustomEditor for CuboidLayout to help debug shader issues
* Null refs from SmoothMotion when keyboard came up


Workspaces/Inspector/Scripts/ListItems/InspectorListItem.cs
```diff
@@ -191,153 +194,159 @@ void ReceiveDrop(BaseHandle handle, object dropObject)
...
-						}
-					}
+			var graphics = clone.GetComponentsInChildren<Graphic>(true);
+			foreach (var graphic in graphics)
+			{
+				graphic.raycastTarget = false;
 
-					m_DragObject = clone.transform;
-					m_ClickedField = null; // Clear clicked field so we don't drag the value
+				if (graphic.GetComponent<Mask>())
```
amirebrahimi: Comment as to why.

Scripts/Interfaces/IFilterUI.cs
```diff
@@ -12,6 +12,25 @@ public interface IFilterUI
 		/// Set accessor for the filter list
 		/// </summary>
 		List<string> filterList { set; }
+		string searchQuery { get; }
+	}
+
+	public static class FilterUIExtensionMethods
+	{
+		public static bool TestFilter(this IFilterUI filterUI, string type)
```
amirebrahimi: Not sure what Test means here. Probably should be named IsFiltered, which means the return value should be reversed.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -53,6 +53,17 @@ public Transform previewOrigin
 		private Transform m_PreviewOrigin;
 
 		/// <summary>
+		/// The transform that the display/preview objects will be parented under
+		/// </summary>
+		public Transform fieldGrabOrigin
+		{
+			get { return m_FieldGrabOrigin; }
+		}
+
```
amirebrahimi: Remove extra line

Scripts/Interfaces/IGetFieldGrabOrigin.cs
```diff
@@ -0,0 +1,18 @@
+﻿#if UNITY_EDITOR
+using System;
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	/// <summary>
+	/// Implementors receive a preview origin transform
```
amirebrahimi: This is incorrect. Probably from copy/paste

Scripts/Interfaces/IGetFieldGrabOrigin.cs
```diff
@@ -0,0 +1,18 @@
...
+	/// </summary>
+	public interface IGetFieldGrabOrigin
+	{
+		/// <summary>
+		/// Get the preview transform attached to the given rayOrigin
```
amirebrahimi: Fix this comment, too.

Scripts/Helpers/ProxyHelper.cs
mtschoen-unity: Ah. I was curious about this. Everything got an extra line in the refactor, presumably because R# does that when it auto-formats. Should I delete these extra lines where I find them?

amirebrahimi: Hold for now until we hear back from formatting group.

PR #97, opened 2017-03-20
Highlighting improvements

### LINKED ISSUE #36: Option to disable Selection Tool highlight
Similar to `EditorUtility.SetSelectedWireframeHidden` in the Editor.

The use being custom highlighting schemes.  In my case I want to highlight the mesh triangles and vertices that are being hovered, but not the object as a whole.

As suggested by Amir [here](https://forum.unity3d.com/threads/selection-tool-still-highlights-objects-after-action-is-consumed.450912/) a workaround in the current state is to simply inherit `ISetHighlight` and manually disable the highlight per-object.

### LINKED ISSUE #96: Selection Tool highlighting issues
The selection tool is highlighting objects even when the ray is hovering over UI elements.

The highlight shader draws on top of everything and is very distracting.

mtschoen-unity:
Closes #96
Closes #36 

This PR swaps our StandardOverlay highlight shader with Valves Silhouette outline shader. It also adds a Transform argument to setHighlight to specify which rayOrigin is doing the highlighting. This way, we can use a different color for the left and right hand. The idea is to make it clear which object will be selected by which ray.

If you specify null for the rayOrigin argument, the HighlightModule uses the DefaultHighlight material, which has a higher render queue for outlining objects in Workspaces.

Because these three highlight materials require their own StencilRef, I increased the minimum StencilRef value to 5 (1 = radial menu, 2 - 4 = highlight, 5+ = workspaces)

This PR also addresses the issue where selection highlights occur when the ray is hovering over UI. I added the IUsesUIBlocking service to check for this in the SelectionTool


Tools/TransformTool/TransformTool.cs
```diff
@@ -227,10 +229,11 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 				if (manipulatorGameObject.activeSelf && hoveringSelection)
 					manipulatorGameObject.SetActive(false);
 
-				foreach (var selection in m_HoverObjects.Values)
+				foreach (var selection in m_HoverObjects)
```
amirebrahimi: Remind me again why we highlight from the transform tool and not the selection tool? We can discuss live, too.

Scripts/Modules/Highlight/HighlightModule.cs
```diff
@@ -7,35 +7,86 @@ namespace UnityEditor.Experimental.EditorVR.Modules
...
+			{
+				set = new HashSet<Transform>();
+				m_NodeMap[node] = set;
+			}
+
+			set.Add(rayOrigin);
 		}
 
-		public void SetHighlight(GameObject go, bool active)
+		public void SetHighlight(GameObject go, Transform rayOrigin, bool active)
```
amirebrahimi: Can we address #36 while we're it as well and provide a method to disable selection highlight for a single object? It could be a part of this method or a separate method. Essentially we would want it to be added to an ignored list or "highlight hidden" list.

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -16,7 +17,8 @@ sealed class SelectionTool : MonoBehaviour, ITool, IUsesRayOrigin, IUsesRaycastR
 
 		public Func<Transform, GameObject> getFirstGameObject { private get; set; }
 		public Transform rayOrigin { private get; set; }
-		public Action<GameObject, bool> setHighlight { private get; set; }
+		public Action<GameObject, Transform, bool> setHighlight { private get; set; }
+		public Action<GameObject> clearHighlight { get; set; }
```
amirebrahimi: This looks unused. Please remove.

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -28,14 +28,16 @@ sealed class SelectionTool : MonoBehaviour, ITool, IUsesRayOrigin, IUsesRaycastR
 
 		public Action<ISetManipulatorsVisible, bool> setManipulatorsVisible { private get; set; }
 
-		public Func<Transform, bool> hoveringUI { private get; set; }
+		public Func<Transform, bool> isHoveringOverUI { private get; set; }
+
+		public Func<Dictionary<Transform, DirectSelectionData>> getDirectSelection { get; set; }
```
amirebrahimi: private get

Scripts/Modules/Highlight/HighlightModule.cs
```diff
@@ -7,60 +8,128 @@ namespace UnityEditor.Experimental.EditorVR.Modules
...
+		Material m_LeftHighlightMaterial;
 
 		[SerializeField]
-		private Material m_HighlightMaterial;
+		Material m_RightHighlightMaterial;
+
+		readonly Dictionary<Material, HashSet<GameObject>> m_Highlights = new Dictionary<Material, HashSet<GameObject>>();
+		readonly Dictionary<Node, HashSet<Transform>> m_NodeMap = new Dictionary<Node, HashSet<Transform>>();
+
+		public event Func<GameObject, Material, bool> onHighlight
```
amirebrahimi: Might as well call this customHighlight. We don't put `on` as a prefix anyway.

Scripts/Core/EditorVR.DirectSelection.cs
```diff
@@ -13,9 +13,10 @@ partial class EditorVR
 		class DirectSelection : Nested
 		{
 			internal IGrabObjects objectsGrabber { get; set; }
-
+			internal Dictionary<Transform, DirectSelectionData> directSelections { get { return m_DirectSelections; } }
+			
 			// Local method use only -- created here to reduce garbage collection
-			readonly Dictionary<Transform, DirectSelectionData> m_DirectSelectionResults = new Dictionary<Transform, DirectSelectionData>();
+			readonly Dictionary<Transform, DirectSelectionData> m_DirectSelections = new Dictionary<Transform, DirectSelectionData>();
```
amirebrahimi: This is no longer just to reduce garbage collections, but now also holds the results, so move it above this area.

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -325,6 +325,14 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
 					linkedObject.isSharedUpdater = IsSharedUpdater;
 				}
 
+				var usesUIBlocking = obj as IIsHoveringOverUI;
```
amirebrahimi: Rename variable to match

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -102,6 +113,11 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 			if (m_EventCamera == null)
 				return;
 
+			// World scaling also scales clipping planes
+			var viewerScale = getViewerScale();
+			m_EventCamera.nearClipPlane = m_OriginalNearClipPlane * viewerScale;
```
mtschoen-unity: I changed this back to copying the values over from the main camera, but kept it here in the code since we didn't need to do it for every intersection check

PR #111, opened 2017-03-24
Fix OnHierarchyChanged spam while MiniWorld is open

mtschoen-unity:
We noticed this a while ago when refactoring how ignore lists worked. We added a feature to only update ignore lists for minworlds and the pixel raycast module when the hierachy changes. But, it turned out that OnHierarchyChanged gets called a lot. Currently, it is called every frame while the MiniWorld is open because of changes to the manipulator.  This change stops the spam.

I would like to know more about why this happens only when MiniWorld is open and because of the Manipulator, but I couldn't find a sensible way to debug the source of the change in managed code. I tracked it down to the manipulator by setting a native breakpoint and reading the name of the transform responsible.


mtschoen-unity:
approved

PR #120, opened 2017-03-29
Snapping

### LINKED ISSUE #62: Snapping

mtschoen-unity:
At long last! Closes #62.

@timoni, here's the tweakable stuff:
* SnappingModule.cs
    * Lines 11 - 25 are all tweakable. Most should be pretty self-explanitory. All of the "break" and "max/min" values are in meters, but they are scaled based on distance to camera for manipulator snapping and viewer scale for direct snapping. Changes to these values should "feel" the same regardless of distance or scale. I'll explain a few specific fields:
        * DirectSurfaceSearchScale - multiply this value by the vector from the center of the object to a face of the bounding box. That's how far we cast a ray from the center to search for an object to snap to. The current value is 1.1x.
        * GroundHeight - We should probably make this variable and settable by a slider on the settings menu. Maybe use the scene bounds as a max/min? Or should you be able to drag and move an object? Anyway, this is the world-space Y value that ground snapping uses
        * WidgetScale - The "Widget" is the little ray that we show when doing surface snapping. It points from the raycast hit out along the surface normal. It is scaled based on distance to the camera like the manipulator, but you can increase or decrease this value to change its relative size.
        * Direct/Manipulator/Ground/Surface - There are different settings for different types of manipulation. It would certainly simplify things if any of these values overlap and can share variables.
    * Default settings are setup in lines 217-221. If we want to change snapping settings on first run, we will change them here. Otherwise, snapping settings are saved on closing the window and loaded when you open it.
* Transformtool.cs
    * LazyFollowTranslate - The speed at which the object will follow the manipulator when translating, whether or not snapping is enabled
    * LazyFollowRotate The speed at which the object will follow the manipulator when rotating, whether or not snapping is enabled
    * DirectLazyFollowTranslate - The speed at which the object's position will follow the controller, when direct manipulating and snapping. Otherwise direct manipulation is 1:1
    * DirectLazyFollowRotate - The speed at which the object's rotation will follow the controller, when direct manipulating and snapping. Otherwise direct manipulation is 1:1

I think we could do a second pass on this feature after a little bit of usability testing focused on visuals. The surface snapping widget is a nice start, but we could definitely do something for direct manipulation like I had with my Gizmos. There are some other ideas I had like trying to use overlaps or sweeps instead of raycasts for surface snapping so that you don't have to line up the center of the object if your snapping to a small object. The UI could use a little love, as well. @timoni was going to get back to me about changing the order of some of the settings, which should be a pretty quick revision when the time comes.

@amirebrahimi, @dunity:
Naturally, you should focus on SnappingModule.cs, SnappingModuleUI.cs, and TransformTool.cs. The SubMenuFace code is taken almost directly from David's branch. The manipulators were modified because I needed to provide the rayOrigin and whether we are using a constrained manipulator to the translate event. Basically I need this info flowed through the Transform tool in order to do snapping.

There are a number of style fixes and convention updates. I didn't modify any files that weren't already being modified to implement the feature. This also fixes the manipulator scale issue that came up after PR #111. I also decided to greatly reduce the scale of the spherehandle while it is in use, so that it does not cover up the object.

The reason I modified  the *Connector.cs files is because they caused the NoEditorVR test to fail as I was trying to check my own modifications.

One topic that bears discussing is whether/where to use an enumerator instead of a foreach loop to enumerate dictionaries. Foreach creates garbage, but the enumerator solution is more verbose and you can potentially forget to dispose it, which I think creates an actual memory leak. R# does warn you about this. I have come up with a solution some time earlier when trying to eliminate garbage across the project: you use a method like this:

```
public static KeyValuePair<TKey, TValue>? ForEachInDictionary<TKey, TValue, TExtra>(Dictionary<TKey, TValue> 
	dictionary, Func<KeyValuePair<TKey, TValue>, TExtra, bool> callback, TExtra extra)
{
	KeyValuePair<TKey, TValue>? result = null;

	var enumerator = dictionary.GetEnumerator();
	while (enumerator.MoveNext())
	{
		if (callback(enumerator.Current, extra))
		{
			result = enumerator.Current;
			break;
		}
	}
	enumerator.Dispose();

	return result;
}
```

This is less than ideal, because you have to rely on returning true within the callback method to break out of the loop, and returning out of the calling method would have to be done with an additional out parameter. Furthermore, using an anonymous method also creates garbage, so you have to avoid closures and declare everything statically or as static references to member methods. Ultimately, until we upgrade our runtime, we'll have to stick with using the "var enumerator" approach where we absolutely need to eliminate garbage. My heuristic is that anything which can happen on a per-frame basis should not allocate any garbage. I was actually able to achieve this a few weeks ago and while I only saved about a 1ms per frame on average, it greatly reduced the frequency of GC-related frame drops.

In this case, I have used var enumerator and other methods to eliminate all GC allocations in the IntersectionModule in order to properly test its performance vs the PixelRaycastModule. Speaking of performance testing. After trying the scene with a couple of 1M triangle dragons, I loaded up Viking Village, which also ran fine, albeit on my 1080.


Scripts/Modules/SnappingModule/Shaders/GroundPlane.shader
```diff
@@ -0,0 +1,217 @@
+﻿Shader "EditorVR/GroundPlane"
```
amirebrahimi: Where does this show up?

Scripts/Modules/SnappingModule/SnappingModuleUI.cs
```diff
@@ -0,0 +1,88 @@
+﻿#if UNITY_EDITOR
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	public class SnappingModuleUI : MonoBehaviour
```
amirebrahimi: Is this the Settings UI? If so, should probably be called SnappingModuleSettingsUI.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+		const float k_ManipulatorSurfaceSnapBreakDist = 0.1f;
+
+		const float k_DirectSurfaceSearchScale = 1.1f;
+		const float k_DirectSurfaceSnapBreakDist = 0.03f;
+		const float k_DirectGroundSnapMin = 0.03f;
+		const float k_DirectGroundSnapMax = 0.07f;
+
+		const float k_WidgetScale = 0.03f;
+
+		const string k_SnappingEnabled = "EditorVR.Snapping.Enabled";
```
amirebrahimi: SnappingModule shouldn't be storing it's own settings directly in EditorPrefs, but you won't have the serialization system until the layout saving PR comes through.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+		GameObject m_Widget;
+
+		[SerializeField]
+		GameObject m_SettingsMenuPrefab;
+
+		SnappingModuleUI m_SnappingModuleUI;
+
+		class SnappingState
+		{
+			public Vector3 currentPosition;
```
amirebrahimi: Public fields aren't allowed, so either autoprop these or turn this into a struct.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+			{
+				Reset();
+				m_DisableAll = !value;
+
+				if (m_SnappingModuleUI)
+					m_SnappingModuleUI.snappingEnabled.isOn = value;
+			}
+		}
+
+		public bool groundSnapping
```
amirebrahimi: groundSnappingEnabled or snapToGround? Not strong on this, but let's discuss.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+			m_Widget = ObjectUtils.Instantiate(m_Widget, transform);
+			m_Widget.SetActive(false);
+
+			groundSnapping = true;
+			surfaceSnapping = true;
+
+			directSnapping = true;
+			manipulatorSnapping = true;
+
+			if (EditorPrefs.HasKey(k_SnappingEnabled))
```
amirebrahimi: Use the overload of GetBool() with default parameter and do for all those below. Can be combined with the initialization that you were doing.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+					return true;
+			}
+
+			position = targetPosition;
+			rotation = targetRotation;
+
+			return false;
+		}
+
+		bool PerformManipulatorSurfaceSnapping(Transform rayOrigin, ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, SnappingState state, Quaternion targetRotation, float breakDistance)
```
amirebrahimi: Instead of Perform is there a verb we can use instead? ManipulatorSnapToSurface?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+			var startRotation = state.startRotation;
+			var upVector = startRotation * Vector3.back;
+			var maxRayLength = k_SurfaceSnappingMaxRayLength * this.GetViewerScale();
+
+			var ray = new Ray(rayOrigin.position, rayOrigin.forward);
+			return PerformSurfaceSnapping(ray, ref position, ref rotation, targetPosition, state, offset, targetRotation, rotationOffset , upVector, m_IgnoreList, breakDistance, maxRayLength)
+				|| TryBreakSurfaceSnapping(ref position, ref rotation, targetPosition, startRotation, state, breakDistance);
+		}
+
+		bool PerformDirectSurfaceSnapping(ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, SnappingState state, Quaternion targetRotation, float breakDistance)
```
amirebrahimi: Same here.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+					return true;
+			}
+
+			if (TryBreakSurfaceSnapping(ref position, ref rotation, targetPosition, targetRotation, state, breakDistance))
+				return true;
+
+			return false;
+		}
+
+		static Vector3 GetDirection(int i)
```
amirebrahimi: Instead of these methods can you create a constant array of a struct that defines the direction, rotation, up vector, etc. and iterate over that instead of having a magic number in your for loops?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+					rotation = targetRotation;
+					state.surfaceSnapping = false;
+				}
+
+				return true;
+			}
+			return false;
+		}
+
+		static void SetupIgnoreList(List<GameObject> ignoreList, Renderer[] playerHeadObjects, GameObject go)
```
amirebrahimi: I don't like that we are referring directly to "player head" in our snapping module. Can you come up with an uncoupled way to do this?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+			ignoreList.Clear();
+			ignoreList.Add(go);
+
+			for (int i = 0; i < playerHeadObjects.Length; i++)
+			{
+				ignoreList.Add(playerHeadObjects[i].gameObject);
+			}
+		}
+
+		bool PerformSurfaceSnapping(Ray ray, ref Vector3 position, ref Quaternion rotation, Vector3 statePosition, SnappingState state, Vector3 boundsOffset, Quaternion targetRotation, Quaternion rotationOffset, Vector3 upVector, List<GameObject> ignoreList, float breakDistance, float raycastDistance)
```
amirebrahimi: Let's discuss naming

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,709 @@
...
+				state.snappedPosition = position;
+				state.snappedRotation = snappedRotation;
+				return true;
+			}
+
+			return false;
+		}
+
+
+		bool PerformGroundSnapping(ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation, SnappingState state, float groundSnapMin, float groundSnapMax)
```
amirebrahimi: Let's discuss naming

Menus/MainMenu/MainMenu.cs
```diff
@@ -97,7 +99,7 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 			consumeControl(mainMenuInput.blockY);
 
 			const float kFlickDeltaThreshold = 0.5f;
-			if ((proxyType != typeof(ViveProxy) && Mathf.Abs(rotationInput) >= kFlickDeltaThreshold && Mathf.Abs(m_LastRotationInput) < kFlickDeltaThreshold)
+			if (proxyType != typeof(ViveProxy) && Mathf.Abs(rotationInput) >= kFlickDeltaThreshold && Mathf.Abs(m_LastRotationInput) < kFlickDeltaThreshold
```
mtschoen-unity: No change to the logic. R# was telling me those parenthesis were redundant. There's an || on the next line, but everything in this line are &&s

Menus/MainMenu/Scripts/MainMenuFace.cs
```diff
@@ -122,48 +129,6 @@ private IEnumerator AnimateVisibility(bool show)
 
 			m_VisibilityCoroutine = null;
 		}
-
-		void BeginVisuals()
```
mtschoen-unity: I did not get a chance to discuss with @dunity, no. I saw that one of the two methods was completely unused, and found that removing the call to EndVisuals did not cause any noiticeable change.  They seem to have been replaced by AnimateVisibility. They were likely an oversight or the result of an automatic merge.

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -88,7 +88,7 @@ internal void UpdateRayForDevice(DeviceData deviceData, Transform rayOrigin)
 			{
 				var mainMenu = deviceData.mainMenu;
 				var customMenu = deviceData.customMenu;
-				if (mainMenu.visible || (customMenu != null && customMenu.visible))
+				if (mainMenu.visible || customMenu != null && customMenu.visible)
```
mtschoen-unity: Same as above. && takes precedence over ||.

Scripts/Core/EditorVR.UI.cs
```diff
@@ -15,7 +15,7 @@ partial class EditorVR
 
 		class UI : Nested, IInterfaceConnector
 		{
-			const byte k_MinStencilRef = 5;
+			const byte k_MinStencilRef = 2;
```
mtschoen-unity: The stencil from the outline shader caused the UI mask on the menu to bleed through. The fix ended up resolving the need for different stencil refs per outline material. Thus, refs 2, 3, and 4 go back to the pool.

Scripts/Core/EditorVR.cs
```diff
@@ -400,14 +416,14 @@ void LateBindNestedModules(IEnumerable<Type> types)
...
-						Nested dependency = null;
+						Nested dependency;
 						if (m_NestedModules.TryGetValue(dependencyType, out dependency))
 						{
 							var map = type.GetInterfaceMap(lateBinding);
 							if (map.InterfaceMethods.Length == 1)
 							{
 								var tm = map.TargetMethods[0];
-								tm.Invoke(nestedModule, new[] { dependency });
+								tm.Invoke(nestedModule, new object[] { dependency });
```
mtschoen-unity: This ended up being benign. I'm going to revert it.

Scripts/Modules/SnappingModule/Shaders/GroundPlane.shader
mtschoen-unity: This is used on the transparent plane that shows up when an object is currently ground snapping. A normal Standard shader or existing custom shader either draws on top of everything or has a lot of z fighting and flicker when the scene has flat ground objects.

Scripts/Core/InterfaceConnectors/ActionsModuleConnector.cs
```diff
@@ -55,3 +56,4 @@ public void DisconnectInterface(object obj)
 		}
 	}
 }
+#endif
```
mtschoen-unity: These have been fixed but for some reason the comments weren't cleared.

mtschoen-unity:
I've addressed the issues we talked about at today's schedule meeting. Pending a final review, this should be good to merge.

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -214,9 +214,17 @@ void OnProxyActiveChanged(IProxy proxy)
 				}
 			}
 
+			internal void UpdateRaycasts()
+			{
+				var intersectionModule = evr.GetModule<IntersectionModule>();
```
amirebrahimi: If this happens every frame, then you might want to cache the intersection module.

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -214,9 +214,17 @@ void OnProxyActiveChanged(IProxy proxy)
...
+				var distance = k_DefaultRayLength * Viewer.GetViewerScale();
+				ForEachRayOrigin(rayOrigin => { intersectionModule.UpdateRaycast(rayOrigin, distance); });
+			}
+
 			internal void UpdateDefaultProxyRays()
 			{
 				var inputModule = evr.GetModule<MultipleRayInputModule>();
```
amirebrahimi: Same here; If this is called every frame, then go ahead and cache these modules.

Scripts/Interfaces/IUsesSnapping.cs
```diff
@@ -0,0 +1,62 @@
...
+	public interface IUsesSnapping
+	{
+	}
+
+	public static class IUsesSnappingMethods
+	{
+		internal delegate bool ManipulatorSnappingDelegate(Transform rayOrigin, GameObject[] objects, ref Vector3 position, ref Quaternion rotation, Vector3 delta);
+		internal delegate bool DirectSnappingDelegate(Transform rayOrigin, GameObject go, ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation);
+
+		internal static ManipulatorSnappingDelegate manipulatorSnapping { get; set; }
```
amirebrahimi: This should be a verb rather than a noun; manipulatorSnap? Adjust the delegate and external method, too

Scripts/Interfaces/IUsesSnapping.cs
```diff
@@ -0,0 +1,62 @@
...
+		internal static DirectSnappingDelegate directSnapping { get; set; }
```
amirebrahimi: directSnap? Adjust the delegate and external method, too.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+				rotatedBounds.center -= position;
+				this.rotatedBounds = rotatedBounds;
+				identityBounds.center -= position;
+				this.identityBounds = identityBounds;
+			}
+		}
+
+		struct SnappingDirection
+		{
+			public Vector3 direction { get; set; }
```
amirebrahimi: Make these all public fields; According to our convention data in structs can only be fields.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+
+				Reset();
+				m_SurfaceSnappingEnabled = value;
+
+				if (m_SnappingModuleSettingsUI)
+					m_SnappingModuleSettingsUI.surfaceSnapping.isOn = value;
+			}
+		}
+
+		public bool pivotSnapping
```
amirebrahimi: pivotSnappingEnabled

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+				}
+			}
+			else
+			{
+				m_GroundPlane.SetActive(false);
+				m_Widget.SetActive(false);
+			}
+		}
+
+		public bool ManipulatorSnapping(Transform rayOrigin, GameObject[] objects, ref Vector3 position, ref Quaternion rotation, Vector3 delta)
```
amirebrahimi: ManipulatorSnap?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+					rotation = targetRotation;
+				}
+			}
+
+			position += delta;
+
+			return false;
+		}
+
+		public bool DirectSnapping(Transform rayOrigin, GameObject go, ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation)
```
amirebrahimi: DirectSnap?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+					return true;
+			}
+
+			if (TryBreakSurfaceSnapping(ref position, ref rotation, targetPosition, targetRotation, state, breakDistance))
+				return true;
+
+			return false;
+		}
+
+		static bool TryBreakSurfaceSnapping(ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation, SnappingState state, float breakDistance)
```
amirebrahimi: TryBreakSurfaceSnap?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+					return true;
+				}
+			}
+
+			return false;
+		}
+
+		SnappingState GetSnappingState(Transform rayOrigin, GameObject go, Vector3 position, Quaternion rotation)
+		{
+			Dictionary<GameObject, SnappingState> states;
```
amirebrahimi: Instead of copy/pasted code, can you keep a cache field for a local array of 1 element that you can set and use the other method below?

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -0,0 +1,780 @@
...
+				}
+
+				return true;
+			}
+			return false;
+		}
+
+		static void SetupIgnoreList(List<GameObject> ignoreList, Renderer[] moduleIgnoreList, GameObject go)
+		{
+			ignoreList.Clear();
```
amirebrahimi: Instead of copy/pasted code, can you keep a cache field for a local array of 1 element that you can set and use the other method below?

Scripts/Modules/SnappingModule/SnappingModuleSettingsUI.cs
```diff
@@ -0,0 +1,88 @@
...
+
+		[SerializeField]
+		Toggle m_DirectSnapping;
+
+		public Toggle snappingEnabled
+		{
+			get { return m_SnappingEnabled; }
+		}
+
+		public Toggle groundSnapping
```
amirebrahimi: This is minor, but you could be consistent and change all of these (e.g. groundSnapping) to xEnabled (e.g. groundSnappingEnabled).

mtschoen-unity:
@amirebrahimi I have addressed your comments

Scripts/Core/EditorVR.cs
```diff
@@ -48,8 +46,13 @@ sealed partial class EditorVR : MonoBehaviour
 		DirectSelection m_DirectSelection;
 		Menus m_Menus;
 		UI m_UI;
+		Interfaces m_Interfaces;
+		MiniWorlds m_MiniWorlds;
```
amirebrahimi: Perhaps I wasn't clear enough in the comments. I only needed you to cache the nested modules locally in the nested module that is making use of these lookups during a frequently called method. It isn't necessary outside of that. The ones here only exist because of their use in Update() or because they have to be spawned prior (e.g. m_Interfaces). Please revert any of the new ones and move them locally to a nested module. It might be easier to simply roll back changes to the specific files you changed related to this change, but I'm fine with you leaving whatever modules were already accessible. I implemented GetNestedModule<> primarily to avoid multiple state variables in EVR, but then someone (possibly dylanu) requested that I cache the ones used in Update. Not even sure if that is actually necessary for performance reasons.

mtschoen-unity: Yes. I misunderstood. However, m_MiniWorlds remains to be necessary because it and m_MultipleRayInputModule are used every frame in ProcessInput.

I have pushed changes that address this feedback. Would you like me to remove more of the references I added to evr.m_SomeModule? I figured that as long as we had them we might as well use them.

Scripts/Core/EditorVR.cs
```diff
@@ -185,25 +178,25 @@ void Awake()
 			m_Interfaces.ConnectInterfaces(intersectionModule);
 			intersectionModule.Setup(spatialHashModule.spatialHash);
 
-			menus.mainMenuTools = tools.allTools.Where(t => !tools.IsPermanentTool(t)).ToList(); // Don't show tools that can't be selected/toggled
+			var snappingModule = AddModule<SnappingModule>();
+			snappingModule.raycast = intersectionModule .Raycast;
```
amirebrahimi: Remove whitespace between intersectionModule and .

Scripts/Interfaces/IUsesSnapping.cs
```diff
@@ -0,0 +1,62 @@
...
+		/// Clear state information for a given ray
+		/// </summary>
+		/// <param name="rayOrigin">The ray whose state to clear</param>
+		public static void ClearSnappingState(this IUsesSnapping usesSnaping, Transform rayOrigin)
+		{
+			clearSnappingState(rayOrigin);
+		}
+	}
+}
+#endif
```
amirebrahimi: NL

PR #123, opened 2017-04-02
Save workspace state + move workspace tool

### LINKED ISSUE #20: Save workspace state
Save the layout of workspaces and possibly even the user's location

### LINKED ISSUE #122: Move workspace tool
This built-in tool allows you to:
1. Move all workspaces together by holding trigger over head for a short period of time and keeping it held down while moving the controller.
2. Close all workspaces by holding trigger over head for a short period of time and while keeping trigger held, throwing them down, and then releasing the trigger.
3. Reset workspace rotation to face viewer by double-trigger press.

amirebrahimi:
Closes #122 
Resolves #20


Workspaces/ProjectWorkspace/Scripts/FolderListViewController.cs
```diff
@@ -18,10 +18,10 @@ sealed class FolderListViewController : NestedListViewController<FolderData, Fol
 		[SerializeField]
 		Material m_ExpandArrowMaterial;
 
-		string m_SelectedFolder;
-
 		readonly Dictionary<string, bool> m_ExpandStates = new Dictionary<string, bool>();
 
+		public string selectedFolder { get; private set; }
```
mtschoen-unity: Should we replace SelectFolder with a setter on selectedFolder? I don't like having both at the same time. If we want to keep setFolder as a method, we should instead use a getFolder method/delegate.

Workspaces/ProjectWorkspace/Scripts/FolderListViewController.cs
```diff
@@ -140,7 +148,12 @@ static FolderData GetFolderDataByGUID(FolderData data, string guid)
 				{
 					var folder = GetFolderDataByGUID(child, guid);
 					if (folder != null)
+					{
+						if (folderToRootCallback != null)
+							folderToRootCallback(folder);
```
mtschoen-unity: Shouldn't this be called on child, not folder? For a > 3 level deep tree, we won't get the root folder to expand.

Workspaces/ProjectWorkspace/ProjectWorkspace.cs
```diff
@@ -144,6 +158,22 @@ public override void Setup()
...
+			preferences.scaleFactor = m_ProjectUI.assetGridView.scaleFactor;
+			preferences.selectedFolder = m_ProjectUI.folderListView.selectedFolder;
+			return preferences;
+		}
+
+		public void OnDeserializeWorkspace(object obj)
+		{
+			var preferences = (Preferences)obj;
+			m_ProjectUI.assetGridView.scaleFactor = preferences.scaleFactor;
+			m_ProjectUI.folderListView.SelectFolder(preferences.selectedFolder);
```
mtschoen-unity: I would recommend that we also save out m_ExpandStates (or at least a list of the guids that are expanded). If a guid changes, we'll lose the expand state, but there shouldn't be any ill effect, as long as we strip serialized expand states that don't have a matching guid at some point.

Scripts/Interfaces/ISerializeWorkspace.cs
```diff
@@ -0,0 +1,10 @@
+﻿using UnityEngine;
```
mtschoen-unity: Also #if UNITY_EDITOR

Scripts/Core/Interfaces/ISerializePreferences.cs
```diff
@@ -0,0 +1,10 @@
+﻿using UnityEngine;
```
mtschoen-unity: Also #if UNITY_EDITOR

amirebrahimi:
@mtschoen-unity - Ready for review again

mtschoen-unity:
Added a few remaining #if UNITY_EDITOR, tested things. We're all good on my end.

PR #127, opened 2017-04-04
Workspace Resizing

### LINKED ISSUE #14: Swap dipping into front with using grip to move workspace
Instead of dipping pointer into the front, use the grip button to move the workspace

### LINKED ISSUE #15: Improve workspace resizing
Adjust colliders to match the frame geometry, so that they can be grabbed easily for resizing instead of having to dip in the area around the workspace.

mtschoen-unity:
Closes #15, closes #14.

The main focus is on WorkspaceUI. I have replaced a lot of the offsets in Update and bounds_set with an offset/scale to the frame blendshape. The offset takes into account the size of the frame when the blend shape weight is 0. By lining up the frame with the bounding box we eliminate most of the discrepancy, and these colliders work for all possible worskspace sizes. As such, there is now a cap on a 100m-wide workspace. Good luck making one that big! (Note that scaling yourself doesn't count. The blendshape is done in local space)

The old 4-collider solution has been replaced with a lot more colliders. I originally had the left/right colliders extend all the way to the front of the frame, but when the workspace is at an angle (like 45 degrees), you actually end up putting the ray all the way inside the collider when you try to grab the angled part of the frame. As such, there is a pair of colliders that match the position of the angled portion, and a lower collider that grows to cover the lower rectangular area that is left. A picture should help:
image

Moving workspaces is now done by moving the tip of either cone anywhere into the frame bounds, pulling the grip trigger and moving your hand.  Resizing the workspace is done by moving the tip of the cone into one of the colliders shown above. All of the handles on those colliders call a hover method on WorkspaceUI which checks for hovers and input and reacts accordingly. Whether to use a single axis or two-axis corner-drag is used by checking the x/z position of the pointer in local workspace space. The corner size, thickness, and frame height are serializable on the workspaceUI prefab. @timoni, this is all tweakable, but it's a lot easier to do if you drop a few lines of code into WorkspaceUI.cs to update the colliders in edit mode while you're tweaking stuff. We can also mess around with a workspace with the colliders visible to fine-tune collision points in a later pass. I know you also wanted to increase the frame thickness, too.

The resize icons crossfade (a little more quickly) in, and reflect whether the resize will be a corner or side resize. Side resize icons smooth follow the controller, and try to get "out of the way, so that the controller model does not block them. The idea is so that if you are focused at the corner of a large workspace, a fixed icon in the center or at the side might not be visible.

Only one hand may grab and move or resize a workspace at a time. The currently visible resize icon fades out, and no hovers are processed until the grab completes. I also disable the smooth follow component while resizing workspaces to fix the "wiggle" which used to happen.

I moved most of the resizing and dragging logic into WorkspaceUI because it grew substantially, and relied on a lot of local variables within the class. Workspace is notified simply via a resize event. Because we're using the grips now, we need a custom action map and ProcessInput method, which we already had in MiniWorlds. MiniWorldWorkspace now just makes use of the standard workspace actionmap and processinput now. As before, because this is a two-handed action map, we need to match the left/right ray origin to the left/right actions, which requires the new ray origin properties on Workspace.  I also found it better to use the tip of the cone, rather than the base, which requires getPointerLength.

The change to ISetHighlight is to fix a build error when trying to debug in VS. However, it looks like it might be some old code from the locking branch. @amirebrahimi, I also see a m_HoveredGameObjects in HierarchyListViewController which I don't really understand the use of. It also uses a force parameter on SetHighlight which isn't actually sent to the module. Should we just remove all of this? I didn't want to do so in case it was there for a good reason.

The change to MultipleRayInputModule is a pretty major fix that should probably go into 0.0.4 which fixes controls losing .isHeld when a ray hovers over UI. We were consuming the select control whenever the ray hovered a UI object, rather than just when the select button was pressed.

When debugging the above issue, I noticed that BlinkLocomotionTool was consuming the blink button every frame unnecessarily, so I removed that.

I noticed a scaling issue in the new grid when resizing the miniworld, so I fixed that.

I switched from Bounds to a Vector3 size in ListViewController because the center was ignored, and this wasn't clear. Since we use both size and extents throughout the codebase, I created members for both which are set when you set the size property.

Some workspaces needed resizing or other tweaks.

Though I was asked to remove the m_WorkspaceModule reference in EditorVR last time around, it is now used directly in ProcessInput, thus should be cached. I added DisconnectInterface to IConnectInterfaces in order to allow the workspace module to clean up after itself and not require code in EditorVR to call DisconnectInterfaces for it.


Scripts/Modules/WorkspaceModule.cs
```diff
@@ -51,12 +52,20 @@ class WorkspaceLayout
 		internal List<IWorkspace> workspaces { get { return m_Workspaces; } }
 		readonly List<IWorkspace> m_Workspaces = new List<IWorkspace>();
 
+		internal List<WorkspaceInput> workspaceInputs { get { return m_WorkspaceInputs; } }
+		readonly List<WorkspaceInput> m_WorkspaceInputs = new List<WorkspaceInput>();
```
dunity: should these readonly lists have the k_ prefix?

Workspaces/MiniWorldWorkspace/MiniWorldWorkspace.cs
```diff
@@ -15,18 +14,22 @@
...
+		// Scales larger or smaller than this spam errors in the console
 		const float k_MinScale = 0.01f;
-		const float k_MaxScale = Mathf.Infinity;
+		const float k_MaxScale = 1e12f;
 
 		// Scale slider min/max (maps to referenceTransform uniform scale)
 		const float k_ZoomSliderMin = 0.5f;
 		const float k_ZoomSliderMax = 200f;
 
+		static readonly Vector3 k_LocatePlayerOffset = new Vector3(0.075f, 0.035f, -0.05f);
```
dunity: static readonlys should be ordered above consts

Workspaces/Base/Workspace.cs
```diff
@@ -82,12 +84,7 @@ public Bounds outerBounds
 		/// <summary>
 		/// If true, prevent the resizing of a workspace via the front and back resize handles
 		/// </summary>
-		public bool preventFrontBackResize { set { m_WorkspaceUI.preventFrontBackResize = value; } }
-
-		/// <summary>
-		/// If true, prevent the resizing of a workspace via the left and right resize handles
-		/// </summary>
-		public bool preventLeftRightResize { set { m_WorkspaceUI.preventLeftRightResize = value; } }
+		public bool preventResize { set { m_WorkspaceUI.preventResize = value; } }
```
mtschoen-unity: Good catch. I also forgot to bring this one up with you. I found that left/right and front/back were always used together, and it simplifies our logic to only use one state. Is there a good reason to control these separately at this point?

Workspaces/Base/Workspace.cs
```diff
@@ -365,6 +240,11 @@ IEnumerator AnimateResetSize()
 				yield return null;
 			}
 		}
+
+		public virtual void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
```
mtschoen-unity: Workspace is currently internal, and doesn't require documentation.

Workspaces/InspectorWorkspace/Scripts/InspectorListViewController.cs
```diff
@@ -1,7 +1,7 @@
 ﻿#if UNITY_EDITOR
-using ListView;
 using System;
 using System.Collections.Generic;
+using ListView;
```
mtschoen-unity: Remove and Sort puts System first

Scripts/Modules/WorkspaceModule.cs
mtschoen-unity: No, they are not static.

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -99,7 +99,7 @@ internal void UpdateMenuVisibilityNearWorkspaces()
 						}
 
 						if (m_WorkspaceModule == null)
```
amirebrahimi: As we discussed, go ahead and get rid of the module caching across the board, except where absolutely needed (e.g. m_Interfaces)

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -211,6 +211,17 @@ void OnProxyActiveChanged(IProxy proxy)
 							m_IntersectionModule.AddTester(tester);
 
 							highlightModule.AddRayOriginForNode(node, rayOrigin);
+
+							var workspaceModule = evr.m_WorkspaceModule;
```
amirebrahimi: How about we turn this into an interface that our Workspace can use. So, instead of getting a single ray origin you get all of them?

Scripts/Core/EditorVR.cs
```diff
@@ -324,6 +328,7 @@ void Update()
 
 		void ProcessInput(HashSet<IProcessInput> processedInputs, ConsumeControlDelegate consumeControl)
 		{
+			m_WorkspaceModule.ProcessInputInWorkspaces(consumeControl);
```
amirebrahimi: Call this ProcessInput

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -51,12 +52,20 @@ class WorkspaceLayout
...
 
 		internal static List<Type> workspaceTypes { get; private set; }
 
-		public bool preserveWorkspaces { get; set; }
+		internal Transform leftRayOrigin { private get; set; }
+		internal Transform rightRayOrigin { private get; set; }
+
+		internal bool preserveWorkspaces { get; set; }
+
+		internal Func<Transform, float> getPointerLength { private get; set; }
```
amirebrahimi: Should this be an interface? Just wondering as we add new things that we manually hook up.

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -127,10 +136,13 @@ internal void CreateWorkspace(Type t, Action<IWorkspace> createdCallback = null)
 		{
 			var cameraTransform = CameraUtils.GetMainCamera().transform;
 
-			var workspace = (IWorkspace)ObjectUtils.CreateGameObjectWithComponent(t, CameraUtils.GetCameraRig(), false);
+			var workspace = (Workspace)ObjectUtils.CreateGameObjectWithComponent(t, CameraUtils.GetCameraRig(), false);
```
amirebrahimi: We can't assume that every workspace is a Workspace.

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -140,6 +152,7 @@ internal void CreateWorkspace(Type t, Action<IWorkspace> createdCallback = null)
 
 			var workspaceTransform = workspace.transform;
 			workspaceTransform.position = cameraTransform.TransformPoint(offset);
+			workspaceTransform.rotation = Quaternion.LookRotation(cameraTransform.forward) * DefaultWorkspaceTilt;
```
amirebrahimi: This is already handled by ResetRotation()

Workspaces/Base/Workspace.cs
amirebrahimi: How come you got rid of the left/right resize handles or did you not update the comment?

Workspaces/Base/Workspace.cs
amirebrahimi: How come Workspace doesn't just implement IProcessInput?

Workspaces/Base/WorkspaceUI.cs
```diff
@@ -444,6 +451,185 @@ IEnumerator Start()
 			yield return null;
 		}
 
+		void AdjustHandlesAndIcons()
```
amirebrahimi: This method looks like a lot of repeated/similar code. Any chance you can simply set up an array of constants/offsets and loop over that to handle this update?

Workspaces/Base/Workspace.cs
dunity: The purpose of the separate front/back left/right resize separation was to allow "2D" workspaces (profiler/console/etc) to only allow for the left/right resize handles to appear, but not the front/back handles.

dunity: how is this case being handled with these changes?

Workspaces/MiniWorldWorkspace/MiniWorldWorkspace.cs
```diff
@@ -219,42 +207,50 @@ void Update()
...
+			var rightGrip = workspaceInput.secondaryRight;
+			if (rightGrip.wasJustPressed && miniWorld.Contains(rightRayOrigin.position))
 			{
 				OnPanZoomDragStarted(rightRayOrigin);
-				consumeControl(miniWorldInput.rightGrab);
+				consumeControl(rightGrip);
 			}
 
-			if (miniWorldInput.leftGrab.isHeld || miniWorldInput.rightGrab.isHeld)
+			if (leftGrip.isHeld || rightGrip.isHeld)
```
mtschoen-unity: There's no reason not to...

I like keeping the logic separated, because that way we don't have to call OnPanZoomDragging in two places, and we semantically separate DragStarted, Dragging, and DragEnded in distinct conditions.

Workspaces/MiniWorldWorkspace/Shaders/Grid.shader
```diff
@@ -33,6 +34,7 @@
 				uniform float2 _GridSpacing;
 				uniform float4 _GridCenter;
 				uniform float _GridScale;
+				uniform float _GridFade;
```
mtschoen-unity: There are some other fixes I'd like to make, as well. You can get some weird issues when  you zoom out really far, pan, and zoom back in due to floating point precision issues. This one was very noticeable whenever you resize the workspace, which is why it's addressed in this PR.

Scripts/Modules/WorkspaceModule.cs
mtschoen-unity: Sure! I've also found myself needing this in lots of places. I'll make a GetPointerLength,  but there could be arguments for just adding it to IUsesDirectSelection.

Workspaces/Base/Workspace.cs
mtschoen-unity: It does. IWorkspace implements ICustomActionMap which implements IProcessInput

Workspaces/Base/WorkspaceUI.cs
mtschoen-unity: I did consider this but nothing that I could think of was any shorter or easier to read. I'd be happy to discuss tomorrow.

Workspaces/Base/Workspace.cs
mtschoen-unity: They were both being set to true in Console and Profiler. There is currently no case for just left/right or top/bottom resizing.

Should we bring it back? The main issue with resizing the profiler/console workspaces is that we aren't changing the size of the render texture or the window size, so it doesn't make sense to resize them in any direction.

Scripts/Core/EditorVR.MiniWorlds.cs
```diff
@@ -137,10 +137,10 @@ internal void UpdateMiniWorlds(ConsumeControlDelegate consumeControl)
 					miniWorldRayOrigin.rotation = referenceTransform.rotation * Quaternion.Inverse(miniWorld.miniWorldTransform.rotation) * originalRayOrigin.rotation;
 					miniWorldRayOrigin.localScale = Vector3.Scale(inverseScale, referenceTransform.localScale);
 
-					var directSelection = evr.m_DirectSelection;
+					var directSelection = evr.GetNestedModule<DirectSelection>();
```
amirebrahimi: I'm not sure this is necessary any more? From what I can tell we are just checking for `!= null`.

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -194,7 +193,7 @@ void OnProxyActiveChanged(IProxy proxy)
 								List<GameObject> prefabs;
 								if (extraData.TryGetValue(rayOriginPair.Key, out prefabs))
 								{
-									var ui = evr.m_UI;
+									var ui = evr.GetNestedModule<UI>();
```
amirebrahimi: Cache this outside of the for loop

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -203,16 +202,13 @@ void OnProxyActiveChanged(IProxy proxy)
 								}
 							}
 
-							if (m_IntersectionModule == null)
-								m_IntersectionModule = evr.GetModule<IntersectionModule>();
-
 							var tester = rayOriginPair.Value.GetComponentInChildren<IntersectionTester>();
 							tester.active = proxy.active;
-							m_IntersectionModule.AddTester(tester);
+							evr.GetModule<IntersectionModule>().AddTester(tester);
```
amirebrahimi: Cache outside of for loop

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -203,16 +202,13 @@ void OnProxyActiveChanged(IProxy proxy)
...
 
 							highlightModule.AddRayOriginForNode(node, rayOrigin);
 
-							var workspaceModule = evr.m_WorkspaceModule;
+							var workspaceModule = evr.GetModule<WorkspaceModule>();
```
amirebrahimi: Cache outside of for loop

Scripts/Core/Interfaces/IGetPointerLength.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
amirebrahimi: #if UNITY_EDITOR

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -7,6 +7,7 @@
 using UnityEditor.Experimental.EditorVR.Workspaces;
 using UnityEngine;
 using UnityEngine.InputNew;
+using Valve.VR;
```
amirebrahimi: Remove

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -6,6 +6,8 @@
 using UnityEditor.Experimental.EditorVR.Utilities;
 using UnityEditor.Experimental.EditorVR.Workspaces;
 using UnityEngine;
+using UnityEngine.InputNew;
+using Valve.VR;
```
amirebrahimi: Remove

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -149,10 +164,20 @@ internal void CreateWorkspace(Type t, Action<IWorkspace> createdCallback = null)
 				workspaceCreated(workspace);
 		}
 
+		internal void ProcessInputInWorkspaces(ConsumeControlDelegate consumeControl)
```
amirebrahimi: How about simply ProcessInput? Since it is in the WorkspaceModule, seems redundant that we would specify InWorkspaces, no?

Workspaces/Base/WorkspaceUI.cs
amirebrahimi: Still desiring that this would be simplified, so let's discuss. Any possibility this could be done w/ design as opposed to code?

Scripts/Core/EditorVR.cs
```diff
@@ -286,7 +288,7 @@ void OnDisable()
 		void Shutdown()
 		{
 			if (m_HasDeserialized)
-				serializedPreferences = GetModule<SerializedPreferencesModule>().SerializePreferences();
+			serializedPreferences = GetModule<SerializedPreferencesModule>().SerializePreferences();
```
amirebrahimi: Fix tab

Scripts/Core/InterfaceConnectors/ActionsModuleConnector.cs
```diff
@@ -18,11 +18,15 @@ public void ConnectInterface(object obj, Transform rayOrigin = null)
 					var toolActions = obj as IActions;
 					if (toolActions != null)
 					{
+						// Delay connecting actions to allow tool / module to initialize first
+						EditorApplication.delayCall += () =>
```
amirebrahimi: Why are all of your tabs messed up? Please format this or tab it correctly.

Tools/VacuumTool/VacuumInput.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+using UnityEngine.InputNew;
+
+// GENERATED FILE - DO NOT EDIT MANUALLY
+namespace UnityEngine.InputNew
+{
+	public class VacuumInput : ActionMapInput {
+		public VacuumInput (ActionMap actionMap) : base (actionMap) { }
+		
+		public ButtonInputControl @secondary { get { return (ButtonInputControl)this[0]; } }
```
amirebrahimi: Instead of naming this `secondary` it should reflect the action, which is to `vacuum`.

Tools/VacuumTool/VacuumTool.cs
```diff
@@ -22,8 +26,8 @@ sealed class VacuumTool : MonoBehaviour, ITool, IStandardActionMap, IUsesRayOrig
 
 		public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
 		{
-			var standardInput = (Standard)input;
-			if (standardInput.action.wasJustPressed)
+			var standardInput = (VacuumInput)input;
```
amirebrahimi: rename to `vacuumInput`

Workspaces/Base/ActionMaps/WorkspaceInput.cs
```diff
@@ -7,9 +7,7 @@ namespace UnityEngine.InputNew
 	public class WorkspaceInput : ActionMapInput {
 		public WorkspaceInput (ActionMap actionMap) : base (actionMap) { }
 		
-		public ButtonInputControl @primaryLeft { get { return (ButtonInputControl)this[0]; } }
-		public ButtonInputControl @primaryRight { get { return (ButtonInputControl)this[1]; } }
-		public ButtonInputControl @secondaryLeft { get { return (ButtonInputControl)this[2]; } }
-		public ButtonInputControl @secondaryRight { get { return (ButtonInputControl)this[3]; } }
+		public ButtonInputControl @secondaryLeft { get { return (ButtonInputControl)this[0]; } }
```
amirebrahimi: I would suggest renaming this to what the action is, potentially `moveResize` or something that doesn't indicate what button is actually being pressed.

amirebrahimi:
Inspector UI is messed up somehow:
image

amirebrahimi:
Additionally, the resizing seems to "catch" where I'm expecting that I can move the workspace around instead. This happens easily with the MiniWorldWorkspace.

amirebrahimi:
I added a missing assignment for VacuumTool to the action map. Make sure to grab latest.

PR #130, opened 2017-04-04
Fix Miniworld Direct manipulation

### LINKED ISSUE #39: Cannot select small objects in MiniWorld
It is difficult to select small objects in MiniWorld.  For example if I zoom out in MiniWorld to see a large portion of my scene and/or to work with large assets, it becomes difficult or impossible to select my MiniWorld avatar to move my position in the world.  

If there could be a mechanism for automatically adjusting the collider sizes of objects in MiniWorld based on MiniWorld scale and the proximity of other objects, that could make the MiniWorld more usable for cases where the size gamut of objects in the scene is diverse.

### LINKED ISSUE #126: Rotating world in MiniWorld w/ two-handed scale breaks direct manipulation
Rotate the world with both hands inside the MiniWorldWorkspace and you'll notice that picking with direct selection no longer works properly.

mtschoen-unity:
Closes #126. Also closes #39

The first commit fixes the rotation issue, but this bug inspired me to look into why we were unable to select small objects. Our method for picking maxDistance in the edge test was flawed. Because we weren't testing the full length of the edge, we were missing small objects with big cones.

We lost highlights in the miniworld when we moved all of the highlight logic to the selection tool because the selection tool is only aware of its own rayorigin. The easiest fix was just to move the highlight logic back to the TransformTool.


Tools/TransformTool/TransformTool.cs
```diff
@@ -10,7 +10,8 @@
 namespace UnityEditor.Experimental.EditorVR.Tools
 {
 	sealed class TransformTool : MonoBehaviour, ITool, ITransformer, ISelectionChanged, IActions, IUsesDirectSelection,
-		IGrabObjects, ICustomRay, IProcessInput, ISelectObject, IManipulatorVisibility, IUsesSnapping
+		IGrabObjects, ICustomRay, IProcessInput, ISelectObject, IManipulatorVisibility, IUsesSnapping, ISetHighlight,
+		IIsHoveringOverUI, IUsesRayLocking
```
amirebrahimi: IIsHoveringOverUI and IUsesRayLocking are not used.

Tools/TransformTool/TransformTool.cs
```diff
@@ -240,6 +242,17 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 				if (manipulatorGameObject.activeSelf && (hoveringSelection || hasObject))
 					manipulatorGameObject.SetActive(false);
 
+				var hovers = new Dictionary<Transform, GameObject>(m_HoverGameObjects);
+				foreach (var kvp in hovers)
+				{
+					var rayOrigin = kvp.Key;
+					if (!directSelection.ContainsKey(rayOrigin))
```
amirebrahimi: How come SelectionTool hovering doesn't cover this? We make use of direct selection in there?

mtschoen-unity: I made a note about this in my PR notes. I moved the selection back to the TransformTool because we would otherwise have to make SelectionTool aware of MiniWorlds.

amirebrahimi: As discussed, let's try for having SelectionTool handle all highlighting and turning TransformTool into an ILinkedTool.

Scripts/Core/EditorVR.Tools.cs
```diff
@@ -104,11 +104,11 @@ internal void SpawnDefaultTools(IProxy proxy)
 					AddToolToDeviceData(toolData, devices);
 
 					toolData = SpawnTool(typeof(TransformTool), out devices, inputDevice);
-					AddToolToStack(deviceData, toolData); // Transform doesn't use input, so is added directly to stack
+					AddToolToDeviceData(toolData, devices);
 					var transformTool = (TransformTool)toolData.tool;
 					if (transformTool.IsSharedUpdater(transformTool))
```
amirebrahimi: How come you are doing this?

Scripts/Core/EditorVR.Tools.cs
```diff
@@ -164,6 +164,10 @@ ToolData SpawnTool(Type toolType, out HashSet<InputDevice> usedDevices, InputDev
 					usedDevices.UnionWith(actionMapInput.GetCurrentlyUsedDevices());
 					InputUtils.CollectDeviceSlotsFromActionMapInput(actionMapInput, ref deviceSlots);
 				}
+				else
```
amirebrahimi: Not quite. The logic changed. It still needs to be a separate `if (usedDevices.Count == 0)` because it's possible that even with an action map assigned that no devices came through from the action map (I think this happened for one IExclusiveMode tool).

PR #131, opened 2017-04-04
Fix UI control blocking

mtschoen-unity:
We were consuming the select control prematurely in MultipleRayInputModule, which caused things that rely on isHeld like the WorkspaceMoveTool to break.

This PR only consumes the select control when the trigger is pressed.


mtschoen-unity:
@sarahhockman this one is pretty easy to test. In previous versions, the Move Workspaces tool would get interrupted if the hand you were using to move them hovered over any of them. In this branch, that should not happen.

mtschoen-unity:
So, to be clear, the fix here isn't for the ray poking through, but interrupting actions, such as direct manipulation and "Move all workspaces," which Amir just added. I can show you what I'm talking about.

PR #133, opened 2017-04-05
Move Snapping EditorPrefs to ISerializePreferences

amirebrahimi:
Note: this is for staging

### DIFF
--- a/Scripts/Modules/SnappingModule/SnappingModule.cs
+++ b/Scripts/Modules/SnappingModule/SnappingModule.cs
@@ -1,5 +1,7 @@
 ﻿#if UNITY_EDITOR
+using System;
 using System.Collections.Generic;
+using UnityEditor.Experimental.EditorVR.Core;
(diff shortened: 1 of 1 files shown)


Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -155,20 +148,82 @@ struct SnappingDirection
...
+			bool m_DisableAll;
+
+			// Snapping Modes
+			[SerializeField]
+			bool m_GroundSnappingEnabled;
+			[SerializeField]
+			bool m_SurfaceSnappingEnabled;
+
+			// Modifiers (do not require reset on value change)
+			[SerializeField]
```
mtschoen-unity: This doesn't preserve the defaults. Without any settings saved, the following should default to enabled:

- ground snapping
- surface snapping
- manipulator snapping
- direct snapping

amirebrahimi: Good find. Fixed now.

PR #137, opened 2017-04-06
Fix dropping objects from project into miniworld

### LINKED ISSUE #121: Dragging objects from Project Workspace to MiniWord
The result of dragging objects from Project Workspace and dropping them in MiniWord is unpredictable. Usually the system prevents dropping them, but sometimes it allows (specially by double-clicking the trigger for dropping). In such situation an scaled version of the object will be placed in world.

Tested using: EditorVR-0-0.0.3/HTC Vive

mtschoen-unity:
Closes #121

A lot of this was just clean up of miniworld interactions that have changed as a result of other features.

The isValid check was hard to track down, but was originally put there because pan/zoom actions were interfering with manipulating things in the miniworld.

I also noticed a snapping bug when using certain assets and fixed it.

### DIFF
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -190,7 +190,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
 			UnityBrandColorScheme.sessionGradient = UnityBrandColorScheme.GetRandomGradient();
 
 			var sceneObjectModule = AddModule<SceneObjectModule>();
-			sceneObjectModule.shouldPlaceObject = (obj, targetScale) =>
+			sceneObjectModule.tryPlaceObject = (obj, targetScale) =>
 			{
 				foreach (var miniWorld in m_MiniWorlds.worlds)
 				{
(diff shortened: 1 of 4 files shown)


Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -532,7 +532,11 @@ void AddToIgnoreList(GameObject[] objects)
 
 			for (int i = 0; i < objects.Length; i++)
 			{
-				m_CombinedIgnoreList.Add(objects[i]);
+				var renderers = objects[i].GetComponentsInChildren<Renderer>();
```
amirebrahimi: Since the snapping module just went in with this release would this be a bug that would go out with 0.0.4?

Scripts/Core/EditorVR.cs
```diff
@@ -190,7 +190,7 @@ void Awake()
 			UnityBrandColorScheme.sessionGradient = UnityBrandColorScheme.GetRandomGradient();
 
 			var sceneObjectModule = AddModule<SceneObjectModule>();
-			sceneObjectModule.shouldPlaceObject = (obj, targetScale) =>
+			sceneObjectModule.tryPlaceObject = (obj, targetScale) =>
```
amirebrahimi: I'm not sure I agree with the rename here semantically. When you look at how it is used in SceneObjectModule, the module is asking whether it should place the object. Now when I read this, it is like TryGetValue where I'm expecting the delegate to try and place the object now. If you want to keep the naming, then we'll need to reverse the logic of where it is used to `if (tryPlaceObject == null || !tryPlaceObject(obj, targetScale))`

Scripts/Modules/SnappingModule/SnappingModule.cs
mtschoen-unity: Yeah. I think that is a good idea.

Scripts/Core/EditorVR.cs
mtschoen-unity: I renamed it because if the object is inside the miniworld, we do actually place the object. That is pretty much all this delegate is responsible for. This was where I had to add code to add the object to the spatial hash.

I have reversed the logic as you described.

PR #140, opened 2017-04-09
Contexts

### LINKED ISSUE #48: Permanent tools (e.g. ILocomotor) cannot be overridden
EditorVR.Tools.cs filters all tools that are ILocomotors as being permanent tools.

This is done at line 42 as follows:

`		bool IsPermanentTool(Type type)
		{
			return typeof(ITransformer).IsAssignableFrom(type)
				|| typeof(SelectionTool).IsAssignableFrom(type)
				|| typeof(ILocomotor).IsAssignableFrom(type)
				|| typeof(VacuumTool).IsAssignableFrom(type);
		}
`

This is very likely a mistake.  You probably meant to explicity filter BlinkLocomotor and your other permanent locomotion tools.

The behavior of the code as you have it above, filters out my tool from showing up on menus, because the system believes all ILocomotor tools are permanent.  Which they're not.  It's perfectly reasonable for me to create a special locomotion tool that I want to be activated from the menu.

### LINKED ISSUE #51: EditorVR Contexts
Originally when we were discussing EditorVR design we knew that we'd want to experiment with every part of the system from menus, UIs, tools, etc. One idea we had at the time is that a group of these might be a complete package from a developer. So, the developer might have the need to re-skin all of EditorVR. This isn't entirely what @leith-bartrich is alluding to in #48, but it has parts of it.

So, details need to be flushed out about what a context encompasses. I suppose without a context that you'd be in VR staring at an empty scene with possibly controllers, but nothing else -- no menus, no tools, no workspaces, etc. It begs the question, what is the UX for switching contexts?

Aside from that, there would still need to be a default context at startup as some people are coming to EditorVR as a user with not much Unity knowledge and not as a developer who is looking to build tools or a custom VR editing environment. So, it would be a non-starter if a user had to code something or wire up some settings in a ScriptableObject on disk or a menu somewhere.

Opening this up for comments.

amirebrahimi:
Closes #51 
Resolves #48 

This is based off of PR #58 from @leith-bartrich.

Currently, this allows you to create multiple editing contexts based off of the same or different VR editors (e.g. EditorVR). For EditorVR, you can set up your default tools in each of these editing contexts, which should address the request from tool devs to swap out the default locomotor or selection tools.

You can swap between contexts via the VRView window or using the sample tool by pressing a button. Primary and secondary triggers cause a push and a pop of the contexts, but this doesn't really work yet without having a VR editor that can suspend and resume properly.

Pushing/popping contexts is still questionable because it requires a good bit of management to suspend the state of the VR editor and to restore the state properly. It's not as easy as disabling/enabling the main GameObject for the VR editor. For now, there would be a good bit of work in EditorVR, so if you try to push a new VR editor on top of EditorVR an exception will be thrown. However, you can swap out editing contexts completely. Try swapping between EditorVR and Empty to see that working.


amirebrahimi:
@leith-bartrich , once you accept the read-only invite, then I can add you as an official reviewer.

Scripts/Interfaces/ISetEditingContext.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	public interface ISetEditingContext
+	{
+		/// <summary>
+		/// All of the available editing contexts
+		/// </summary>
+		List<IEditingContext> allContexts { set; }
```
leith-bartrich: Probably should be a delegate such that it can be enumerated dynamically?  Also, probably should explicitly be IShellContextLaunchers or maybe even a IUnitySceneEditorSubContextLauncher?  Which would make a pretty nice plugin system where you can have SubContexts that are marked for use in your Scene Editor rather than just all of them at all times.

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -0,0 +1,85 @@
...
+using UnityEngine.InputNew;
+
+namespace UnityEditor.Experimental.EditorVR.Tools
+{
+	[MainMenuItem("Change Context", "Create", "Push and pop editing contexts")]
+	//[MainMenuItem(false)]
+	sealed class ChangeEditingContextTool : MonoBehaviour, ITool, ICustomActionMap, ISetEditingContext, ITooltip, 
+		ISetTooltipVisibility, ITooltipPlacement, IUsesRayOrigin
+	{
+		[SerializeField]
```
leith-bartrich: Probably should work with the stack rather than the allAvailable in the manager.  Probably should be populated by an enumeration of available IUnitySceneEditorSubContextLaunchers.  Which in turn should provide a "name" and "icon."

Also, PopContext should probably be moved to its own tool, which is always active.  Such that you can always pop out with the reserved menu button.

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -0,0 +1,85 @@
+﻿#if UNITY_EDITOR
+using System;
+using System.Collections.Generic;
+using UnityEditor.Experimental.EditorVR.Core;
```
dunity: remove core

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -0,0 +1,85 @@
+﻿#if UNITY_EDITOR
+using System;
```
dunity: remove system

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -0,0 +1,85 @@
...
+			m_TooltipTransform.position = rayOrigin.position + Vector3.up * 0.1f;
+			var rotation = MathUtilsExt.ConstrainYawRotation(rayOrigin.rotation);
+			m_TooltipTransform.rotation = rotation;
+
+			var change = changeContextInput.change;
+			if (change.negative.wasJustPressed)
+				m_CurrentContextIndex = Mathf.Max(m_CurrentContextIndex - 1, 0);
+			else if (change.positive.wasJustReleased)
+				m_CurrentContextIndex = Mathf.Min(m_CurrentContextIndex + 1, allContexts.Count - 1);
+			consumeControl(change);
```
dunity: add new line

Scripts/Core/Contexts/EditingContextLauncher.cs
```diff
@@ -0,0 +1,227 @@
...
+		List<IEditingContext> m_AvailableContexts = new List<IEditingContext>();
+		string[] m_ContextNames;
+		int m_SelectedContextIndex;
+
+		IEditingContext defaultContext
+		{
+			get
+			{
+				var context = m_AvailableContexts.First();
+
```
dunity: remove extra line; context is applicable to the same if loop that defaultContextName is

Scripts/Core/Contexts/EditingContextLauncher.cs
```diff
@@ -0,0 +1,227 @@
...
+				return context;
+			}
+			set
+			{
+				EditorPrefs.SetString(k_DefaultContext, value.name); 
+			}
+		}
+
+
```
dunity: Remove extra line

Scripts/Core/Contexts/EditingContextLauncher.cs
```diff
@@ -0,0 +1,227 @@
...
+				InputSystem.ExecuteEvents();
+				managers = Resources.FindObjectsOfTypeAll<InputManager>();
+
+				if (managers.Length == 0)
+				{
+					typeof(InputSystem).TypeInitializer.Invoke(null, null);
+					managers = Resources.FindObjectsOfTypeAll<InputManager>();
+				}
+			}
+			Assert.IsTrue(managers.Length == 1, "Only one InputManager should be active; Count: " + managers.Length);
```
dunity: add new line above after if statement

Scripts/Core/Contexts/EditingContextLauncher.cs
```diff
@@ -0,0 +1,227 @@
...
+
+			s_InputManager = managers[0];
+			s_InputManager.gameObject.hideFlags = ObjectUtils.hideFlags;
```
dunity: perhaps cache off s_InputManager.gameObject for use on 216 + 217

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -0,0 +1,85 @@
...
+		{
+			m_TooltipTransform = ObjectUtils.CreateEmptyGameObject(null, transform).transform;
+		}
+
+		void OnDestroy()
+		{
+			if (gameObject.activeInHierarchy)
+				this.HideTooltip(this);
+
+			ObjectUtils.Destroy(m_TooltipTransform.gameObject);
```
dunity: Destroying the toolTip transform via selecting the tool in the main menu, then selecting it again in the main menu (without using the tool), will throw an exception in MathUtilsExt, SetTransformOffset();

Scripts/Core/Contexts/EditingContextLauncher.cs
amirebrahimi: The defaultContextName block is only for whether there is a default context saved. I don't think that the context var initialization is part of that. Happy to discuss if you don't agree.

amirebrahimi:
@dunity - Addressed your comments
@leith-bartrich - Replaced context stack with single context; Added project and user settings (project settings will override if they are updated/newer)

leith-bartrich:
Removal of context stack looks good to me.

Only one hang-up.

You probably still want some kind of easily triggered "back" functionality.  Which allows jumping back to the last context.  To foster the idea of switching contexts quickly and easily.  Maybe it's just a trailing list of N number of contexts entered.  Such that you have the option of issuing a "BackToLastContext" command to the manager.  And it in turn it can just pop the index off that trailing list and enter the given context.  Like the "last channel" button on your TV remote with a longer memory.

It's of course up to the current context how it might expose the ability to issue such a "back" command.

That way, I'm incentivized to push them to your editing context for things that your editing context is good at, because I know they can get back easily, assuming your context has an easy to push "back" button of sorts.  And likewise, I can be a good "community" member by making an easily used "back" button in my context as well.

leith-bartrich:
RestorePreviousEditingContext looks good to me.  Thanks.

amirebrahimi:
Good point @leith-bartrich -- Added the previous list, too.

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -0,0 +1,282 @@
...
+				var context = AssetDatabase.LoadMainAssetAtPath(assetPath) as IEditingContext;
+				availableContexts.Add(context);
+			}
+
+			return availableContexts;
+		}
+
+		internal static string[] GetEditingContextNames()
+		{
+			var availableContexts = EditingContextManager.GetEditingContextAssets();
```
dunity: Remove EditingContextManager

Editor/EditingContextManagerEditor.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
```
dunity: remove System.Collections.Generic;

Editor/EditingContextManagerEditor.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq;
```
dunity: remove linq

Scripts/Core/Contexts/EditingContextManagerSettings.cs
```diff
@@ -0,0 +1,20 @@
+﻿#if UNITY_EDITOR
+using System;
+using UnityEngine;
+using Object = UnityEngine.Object;
```
dunity: remove object

Tools/Temp/ChangeEditingContextTool.cs
```diff
@@ -1,7 +1,6 @@
 ﻿#if UNITY_EDITOR
-using System;
 using System.Collections.Generic;
-using UnityEditor.Experimental.EditorVR.Core;
+using UnityEditor.Experimental.EditorVR.Extensions;
```
dunity: remove extensions using

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -0,0 +1,282 @@
...
+		int m_SelectedContextIndex;
+
+		IEditingContext m_CurrentContext;
+		readonly List<IEditingContext> m_PreviousContexts = new List<IEditingContext>();
+
+		IEditingContext defaultContext
+		{
+			get
+			{
+				var context = m_AvailableContexts.First();
```
mtschoen-unity: Without having picked a default, I'll be loaded into the blink context. Without somehow hardcoding EditorVR to set itself up as a default, I would like to preserve the "first run experience" of starting in the EditorVR context.

I see an interesting opportunity here: Create a _Tutorial context that we can count on to be sorted first, which introduces the user to  EVR and as a part of the tutorial sets up EVR as the default context. In any case, we could just rename Blink to OnlyBlink and rely on the fact that files sort alphabetically and OnlyBlink comes after EditorVR.

Otherwise, I wonder if there is a more clever way to default to EditorVR when you first set up the project.

Tools/Temp/ChangeEditingContextInput.asset
```diff
@@ -0,0 +1,90 @@
+%YAML 1.1
```
mtschoen-unity: It looks like you need to serialize this file. I don't see a "Previous" control in the asset, but I see it in the script. I get errors when I  try to use the tool as a result.

Scripts/Core/Contexts/EditingContextManager.cs
amirebrahimi: I've addressed your point of having a default context. This would also allow us to have a Tutorial context when that time comes.

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -0,0 +1,282 @@
...
+using System.Linq;
+using UnityEditor.Experimental.EditorVR.Utilities;
+using UnityEngine;
+using UnityEngine.Assertions;
+using UnityEngine.InputNew;
+using System.IO;
+
+namespace UnityEditor.Experimental.EditorVR.Core
+{
+	class EditingContextManager : MonoBehaviour
```
amirebrahimi: I'd prefer not to rename it because it sits within the EditorVR.Core namespace and I think is clear from that.

Tools/Temp/ChangeEditingContextInput.asset
amirebrahimi: Fixed now; Editor probably crashed before saving project.

amirebrahimi:
@mtschoen-unity - I believe I've addressed the changes you requested; Regarding the bugs for switching to contexts that don't contain specific tools, we'll need to address those more systematically with dependency tracking. I'll be removing the empty/blink contexts anyway and marking the ChangeContextTool as hidden (it's under the Temp/ folder anyway)

PR #141, opened 2017-04-11
0.0.4 Hotfix

mtschoen-unity:
Hotfixes for 0.0.4 release

### DIFF
--- a/Scripts/Modules/LockModule/LockModule.cs
+++ b/Scripts/Modules/LockModule/LockModule.cs
@@ -47,7 +47,13 @@ namespace UnityEditor.Experimental.EditorVR.Modules
 
 		public bool IsLocked(GameObject go)
 		{
-			return go && (go.hideFlags & HideFlags.NotEditable) != 0;
+			if (!go)
+				return false;
+
+			if (go.transform.IsChildOf(transform)) // EditorVR objects (i.e. PlayerHead) are hidden but not locked
+				return false;
+
+			return (go.hideFlags & HideFlags.NotEditable) != 0;
 		}
 
 		bool ToggleLocked()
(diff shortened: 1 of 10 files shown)


Scripts/Modules/LockModule/LockModule.cs
```diff
@@ -47,7 +47,13 @@ void Awake()
 
 		public bool IsLocked(GameObject go)
 		{
-			return go && (go.hideFlags & HideFlags.NotEditable) != 0;
+			if (!go)
+				return false;
+
+			if (go.transform.IsChildOf(transform)) // EditorVR objects (i.e. PlayerHead) are hidden but not locked
```
amirebrahimi: This doesn't seem necessary. If the NotEditable flag is not set on EditorVR objects, then it shouldn't be locked.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -725,40 +725,40 @@ void SetupUI()
 			handle.hoverEnded += (baseHandle, data) => { text.text = m_Preferences.disableAll ? "Snapping disabled" : "Snapping enabled"; };
 
 			var groundSnappingUI = m_SnappingModuleSettingsUI.groundSnappingEnabled;
-			groundSnappingUI.isOn = groundSnappingEnabled;
+			groundSnappingUI.isOn = m_Preferences.groundSnappingEnabled;
```
amirebrahimi: I don't see how this fixes anything? The properties already routed to m_Preferences.

Tools/TransformTool/TransformTool.cs
```diff
@@ -281,6 +281,10 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 						if (objectGrabbed != null)
 							objectGrabbed(hoveredObject);
 
+						// Only add to selection, don't remove
```
amirebrahimi: I see you moved this back over here, but why?

PR #142, opened 2017-04-12
Fix head grab

amirebrahimi:
Main issue was it was using rayOrigin and not directRayOrigin


amirebrahimi:
@mtschoen-unity - made some changes; up for your review.

PR #145, opened 2017-04-12
Send mouse events to console and profiler workspaces

### LINKED ISSUE #12: Forward ray input as GUI input to standard EditorWindows that show in EditorVR
Once a ray is intersecting with a workspace that has an EditorWindow (see #11) forward any button presses as mouse clicks.

mtschoen-unity:
Closes #12 

Pretty straightforward. I considered tackling the "generic capture window" at the same time, but this was clean enough to just do in one go. Obviously the duplicated baseHandle setup will be unified when we handle this feature.

The one gotcha was brought up in our slack channel as I encountered it. You need to send an OS message to dismiss the context menu which is used for dropdowns in the profiler (like the Active Profiler dropodown). By design, you can send the message right before the menu pops up to pre-emptively close it, hence we do not need to sleep the thread in case the menu is delayed in opening. However, the thread is necessary because SendMessage causes a hitch (I assume a context switch or something) if we call it from the main thread.

I also have to "filter" the click positions, rejecting any positions with y values less than 25, which I determined empircally, but seems to be roughly the size of the "draggable" region of the window which is not actually rendered. This is to fix an issue where I could sometimes send mouse positions to the drag region and move the window slightly, which causes the rendering to fail. This also seems to be different for custom editor windows, so we might have to revisit this when generalizing the window capture feature.


Scripts/Helpers/EditorWindowCapture.cs
```diff
@@ -99,6 +117,60 @@ private void Update()
...
+			var contentSize = new Vector2((worldCorners[1] - worldCorners[2]).magnitude,
+				(worldCorners[0] - worldCorners[1]).magnitude);
+			localPosition.x /= contentSize.x;
+			localPosition.z /= -contentSize.y;
+			var rectPosition = new Vector2(localPosition.x + 0.5f, localPosition.z + 0.5f);
+
+			var rect = m_Window.position;
+			var clickPosition = Vector2.Scale(rectPosition, rect.size) + offset;
+
+			const int windowHeaderSize = 25;
```
amirebrahimi: Let's figure out what is causing the additional offset between setting a mouse position here and sending the event.

Workspaces/ProfilerWorkspace/ProfilerWorkspace.cs
```diff
@@ -45,58 +36,9 @@ bool inView
...
-
-			var bounds = contentBounds;
-			var size = bounds.size;
-			size.z = 0.1f;
-			bounds.size = size;
-			contentBounds = bounds;
-
 			UnityEditorInternal.ProfilerDriver.profileEditor = false;
 
 			m_CaptureWindowRect = GetComponentInChildren<EditorWindowCapture>().GetComponent<RectTransform>();
```
amirebrahimi: Where is your call to base.Setup()?

mtschoen-unity: Not sure why this didn't get hidden, but it's there now.

amirebrahimi:
@mtschoen-unity - Same here; Go ahead and update to latest dev to resolve conflicts and let's continue this review on 5.6.

PR #146, opened 2017-04-13
Make outlines visible in miniworld

### LINKED ISSUE #113: Highlight looks incorrect in MiniWorld
Happens in 0.0.3

If an object is being hovered with the ray, it shows an odd looking highlight in the MiniWorld.

mtschoen-unity:
Closes #113

I didn't think a design review would be required. The miniworld now looks like the world-world. The major change is that we specify the "RenderType" tag to use for replacing shaders, and add silhouette subshaders to the CustomClipPlanes shader.

I also removed the nolighting settings from the clip planes shader to see what it looked like. I didn't see a huge difference in performance (something like ~0.4ms), even with the entire Polygon Adventure demo scene zoomed way out and visible in the miniworld at the same time. I was thinking that we could include a copy of this shader file with the lighting stripped and maybe no alpha or something that users could swap in.

The copy-pasted code is regrettable, but I couldn't get it to work with a cginclude.


amirebrahimi:
@mtschoen-unity - Since there are conflicts, I'd say go ahead and merge latest development and we'll continue this review on 5.6.

PR #150, opened 2017-04-17
0.0.5 staging -> master

amirebrahimi:
Up for review.

[Changes in this version](https://github.com/Unity-Technologies/EditorVR/milestone/4?closed=1)


amirebrahimi:
Thanks @sarahhockman - @mtschoen-unity is going to try and address #159, so please stand-by today to verify.

amirebrahimi:
@sarahhockman - @mtschoen-unity looked at the bug and it is a more involved fix, so since this is not a blocker we're not going to hold 0.0.5. If there aren't any other blockers, then looking for your approval :)

sarahhockman:
@amirebrahimi Got it. 
Other than that, looks good!

PR #154, opened 2017-04-18
Serialize only when preserveLayout is set

### LINKED ISSUE #153: Exception occurs if "Preserve Layout" is not set in EditorVR preferences
https://forum.unity3d.com/threads/using-the-editorvr-without-vr-suggestions.461566/#post-3036820

amirebrahimi:
Fixes #153


amirebrahimi:
@mtschoen-unity - changed it to preserve the previous preferences now

PR #169, opened 2017-05-04
MiniWorld Transform Improvements

### LINKED ISSUE #167: Moving objects in MiniWorld then pressing the trigger
Moving objects in MiniWorld, then pressing the trigger (to select other objects in the main world, deselecting all objects, ...) causes the first object that is moved in MiniWorld to disappear. In fact the scale of the first moved object will become (0, 0, 0). In addition, the trigger of the controller which moved the objects will become disabled.

I experienced this issue in 0.0.5 using HTC Vive.

mtschoen-unity:
Closes #167 

The original bug was caused by the change we made to SelectionTool and TransformTool in order to remove selection logic from TransformTool. In doing so, we gave SelectionTool input priority over TransformTool, and as a result, we were missing the directSelectInput.wasJustReleased which "cleaned up" miniworld transform states. As such, the next trigger press that wasn't blocked by SelectionTool would trigger the cleanup for that MiniWorld interaction at a later time, which sometimes deleted the object or did weird stuff.

I took this opportunity to entirely remove input from UpdateMiniWorlds. It should now in theory not require processing in ProcessInput, but it does matter that we call UpdateMiniWorlds before processing input on the Transform and Selection tool, so I left it where it was.

I've also fixed the interaction slightly so that object scale is always preserved. In the old version, an object which starts outside the MiniWorld, goes inside, and comes back out will be placed back in the real world at a different scale than it started at. This is no longer the case.


mtschoen-unity:
I'm still working in some edge cases that I discovered since posting this PR, so I'm going to close it until I'm finished.

PR #175, opened 2017-05-10
5.6 changes

### LINKED ISSUE #171: VRView window darker than SceneView or HMD
Prior to 5.6, the VRView would render darker than what is actually shown on the HMD due to color space and SRGB not being respected.

### LINKED ISSUE #172: Single pass support
In the 5.4 custom editor build, single pass does not work with EditorVR. With the 5.6 update this will be fixed with one known issue: MiniWorld rendering is currently broken in single pass.

### LINKED ISSUE #173: Error: The compiler this script was imported with is not available anymore
This message would frequently print the Console window during re-compiles. This will be fixed with the upcoming 5.6 release and a better error message will exist for when this error does occur.

### LINKED ISSUE #174: Ctrl+E doesn't launch EditorVR window
After re-compiles, hitting ctrl+e would launch other windows improperly. This has been fixed with the 5.6 release.

amirebrahimi:
Fixes #171 
Closes #172 
Fixes #173 
Fixes #174

This relies on a custom editor build, which can be downloaded [here](http://beta.unity3d.com/download/449bc785cb05/UnityDownloadAssistant-5.6.0xEditorVR-p3.exe).
[Full download page](http://beta.unity3d.com/download/449bc785cb05/download.html)


Scripts/Core/VRView.cs
```diff
@@ -266,7 +231,7 @@ public void CreateCameraTargetTexture(ref RenderTexture renderTexture, Rect came
 
 			if (renderTexture == null)
 			{
-				renderTexture = new RenderTexture(0, 0, 24, format);
+				renderTexture = new RenderTexture(0, 0, 24, format, RenderTextureReadWrite.Linear);
```
mtschoen-unity: Should we add a comment about this fixing the dark mirror?

PR #212, opened 2017-06-05
Play mode switching fix

amirebrahimi:


mtschoen-unity:
What does this fix, exactly? From what I see, it just moves the handling of the editor pref into the context manager?

Also, since we have this and saving of layout, should we add a play button to the menu?

amirebrahimi:
It moves the methods over, but the main thing is that EditorApplication.playmodeStateChanged was not getting called. In re: to adding a play button - that relies on having an action menu, which doesn't exist right now. As a stop-gap we could add play to the radial menu, since it is an action.

mtschoen-unity:
Got it. Looks good to me.

I don't think play makes much sense on the radial menu, so let's hold off.

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -89,6 +94,32 @@ static void EditProjectSettings()
 			Selection.activeObject = settings;
 		}
 
+		// Life cycle management across playmode switches is an odd beast indeed, and there is a need to reliably relaunch
+		// EditorVR after we switch back out of playmode (assuming the view was visible before a playmode switch). So,
+		// we watch until playmode is done and then relaunch.  
+		static void ReopenOnExitPlaymode()
+		{
+			bool launch = EditorPrefs.GetBool(k_LaunchOnExitPlaymode, false);
```
mtschoen-unity: var

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -89,6 +94,32 @@ static void EditProjectSettings()
...
+					EditorApplication.delayCall += ShowEditorVR;
+			}
+		}
+
+		private void OnPlaymodeStateChanged()
+		{
+			if (EditorApplication.isPlayingOrWillChangePlaymode)
+			{
+				EditorPrefs.SetBool(k_LaunchOnExitPlaymode, true);
+				VRView view = VRView.activeView;
```
mtschoen-unity: var

amirebrahimi:
@mtschoen-unity - changes are up

PR #216, opened 2017-06-06
Remove thresholds for movement/rotation speed

### LINKED ISSUE #198: Remove fast/slow translation/rotation thresholds for joystick movement
Particularly rotation. Simply unusable. Can we switch to new Vive version? Say hold down joystick to enter locomotion mode, then hold down trigger and point to move?

amirebrahimi:
Resolves #198 

(including @mtschoen-unity, since you had opinions about this -- we did test this as a group, including @timoni)


mtschoen-unity:
Yes. I did this as well in my dev/schoen/locomotion branch. I didn't quite finish but I do intend to open a PR. Approving this as an interim fix.

PR #221, opened 2017-06-16
Snapping update

### LINKED ISSUE #192: Snapping isn't working with axis/plane movement (both platforms)
Works for direct manipulation and free select circle, but not with axis or plane movement.

### LINKED ISSUE #193: Free movement sphere + snapping leads to very unexpected behavior
Eg, moving an object slightly pushes it forward and backwards tens of meters. I say cut for now or severely limit how much the object can actually move per degree of controller movement.

### LINKED ISSUE #195: Snap to ground not working (both platforms)

mtschoen-unity:
Timoni noticed a couple of unexpected behaviors when testing the snapping feature.

Closes #192, Closes #195, Closes #193 

The issue with the sphere manipulator was actually intended behavior, but we decided that it was best to switch "Limit snapping radius" to be enabled by default, so that objects won't appear to "fly away" unexpectedly.

The linear mainipulator handles simply didn't have a snapping solution, so that was added. Snap to ground was implemented, but too easy to "punch through."

In testing these changes, we discovered a need to add a new "punch through" requirement to the design spec. On top of the rule that it should require more distance break snapping than to snap in the first place, it should also require a much greater break distance if you are moving against the surface normal. This "punch through" break distance is now set to 5x the normal break distance.

In the process, I ended up cleaning up the snapping code a little bit. Because this "punch through" is such a different approach, it is possible we have created new edge cases, so thorough testing is recommended.


Scripts/Interfaces/Entity/IManipulator.cs
```diff
@@ -13,7 +21,7 @@ public interface IManipulator
 		/// Delegate that processes the translation, using the vector3 passed in
 		/// Caller also provides the ray origin that is doing the action, and whether or not this translation is axis-constrained
```
amirebrahimi: Probably want to update the comments to reflect the change from bool to ConstrainedAxis. Something to the effect of "and constrained axes"

Manipulators/StandardManipulator.cs
```diff
@@ -67,7 +67,16 @@ void Update()
 
 		void OnTranslateDragging(BaseHandle handle, HandleEventData eventData)
 		{
-			translate(eventData.deltaPosition, eventData.rayOrigin, !(handle is SphereHandle));
+			ConstrainedAxis constraints = 0;
+			var linear = handle as LinearHandle;
```
amirebrahimi: Create an internal interface called IAxisConstraints that exposes the property, so you can simply check for that to get the constraints off instead of checking for specific classes.

Scripts/Interfaces/Entity/IManipulator.cs
```diff
@@ -4,6 +4,14 @@
 
 namespace UnityEditor.Experimental.EditorVR
 {
+	[Flags]
+	public enum ConstrainedAxis
```
amirebrahimi: Move to a separate file

mtschoen-unity:
@amirebrahimi I've addressed your comments and added one final touch: scaling the ground plane based on viewer scale. I was noticing that you can see the edges when you scale yourself very big. The problem with just using a larger scale all the time is that it flickers at regular scales with a scale of 1000000

Scripts/Core/Interfaces/IAxisConstraints.cs
```diff
@@ -0,0 +1,9 @@
+﻿#if UNITY_EDITOR
+namespace UnityEditor.Experimental.EditorVR
+{
+	interface IConstrainedAxis
```
amirebrahimi: Please rename to IAxisConstraints

Scripts/Data/ConstrainedAxisFlags.cs
```diff
@@ -0,0 +1,17 @@
+﻿#if UNITY_EDITOR
+using System;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	/// <summary>
+	/// Which axes are constrained
+	/// </summary>
+	[Flags]
+	public enum ConstrainedAxis
```
amirebrahimi: Either rename this enum to ConstrainedAxisFlags or the file to ConstrainedAxis.cs.

Scripts/Modules/SnappingModule/SnappingModule.cs
```diff
@@ -511,65 +526,201 @@ public bool DirectSnap(Transform rayOrigin, Transform transform, ref Vector3 pos
...
 				return true;
 
 			return false;
 		}
 
-		static bool TryBreakSurfaceSnap(ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation, SnappingState state, float breakDistance)
+		static bool TryBreakSnap(ref Vector3 position, ref Quaternion rotation, Vector3 targetPosition, Quaternion targetRotation, SnappingState state, float breakDistance)
 		{
-			if (state.surfaceSnapping)
+			if (state.snapping)
```
AndrewTHEManeri: Would really fast motion that stayed within snapping range possible trigger a snap break?

Tools/TransformTool/TransformTool.cs
```diff
@@ -537,12 +539,22 @@ public void DropHeldObjects(Node node)
...
 		{
-			if (constrained)
-				m_TargetPosition += delta;
-			else
-				this.ManipulatorSnap(rayOrigin, Selection.transforms, ref m_TargetPosition, ref m_TargetRotation, delta);
+			switch (constraints)
+			{
+				case ConstrainedAxis.X | ConstrainedAxis.Y:
+				case ConstrainedAxis.Y | ConstrainedAxis.Z:
+				case ConstrainedAxis.X | ConstrainedAxis.Z:
```
AndrewTHEManeri: Should the axis constraints be adjusting the delta here to knock out the constrained/non-constrained axes?

mtschoen-unity: No. This happens in the handle itself. The reason we are doing this is that we don't actually have a solution for snapping the planar handles yet. My original approach was just to treat them as a snap in one axis, then a snap in the other, but that felt unpredictable. When I discussed this with Timoni, we agreed that since the planar handles already "snap" to a plane, it is not as important that they snap to surfaces.

PR #225, opened 2017-06-21
Locomotion Update

### LINKED ISSUE #203: Need a way to reset to default location/undo worldscale movement

mtschoen-unity:
This is a major overhaul of the locomotion tool. First off, I renamed it from BlinkLocomotionTool to LocomotionTool. One of the primary motivations is to clear locomotion off of the thumbsticks  entirely, because joystick locomotion is often overridden by menus (both main menu and radial menu use the joystick) or tools (i.e. Annotation tool). With some distance, I might want joystick locomotion back, but I would want a clear way of indicating to the user whether the joystick is going to move the player or do another action. LocomotionTool.cs is where most of the changes are.

The current methods of locomotion are:
- Blink Locomotion
    - Bound to Menu/B/Y button (Action 2)
    - Toggled in Settings Menu (Fly / Blink)
    - Same as always, with a scale-related visual bugfix (See below)
- Fly (Superman?) Locomotion
    - Bound to Menu/B/Y button (Action 2)
    - Toggled in Settings menu (Fly / Blink)
    - Point the ray in the direction you would like to fly
    - Pull the trigger to increase speed (analog)
    - A/X Button (Action 1) flies in reverse (Rift-only)
    - Either fly button (Action 1 / 2) + Grip (Trigger 2) to rotate world
- Crawl Locomotion
    - Pull grip to "grab" the world and move it around
- Two-handed Scale (Locomotion)
    - Pull both grips (Trigger 2) to move, scale, and rotate the world
    - Fancy visuals
    - Press both thumbsticks/thumbpads (Thumbstick Press) while holding both grips (Trigger 2) to reset scale, keeping your feet at about the same height (same as before)
    - Pull both triggers (Trigger 1) while holding both grips (Trigger 2) to reset position, rotation, and scale to origin/identity (new)
    
Multi-select has been moved from the grip trigger/button (Trigger 2) to the thumbstick/thumbpad press (Thumbstick Press). This is to avoid conflict with the Crawl locomotion, which is always active. To allow this to work on Vive, an ActivationThreshold was added to the RadialMenu which ignores joystick vectors whose magnitude are less than 0.5, to allow some wiggle room on where you can press the thumbpad.

The scale fix for blink visuals was just an issue of scale not being set properly when loading up saved viewer settings, where the blink visuals, which didn't use the IUsesViewerScale interface, thought scale was still set to 1.

The Locomotion Settings are added to the menu via a new interface, ISettingsMenuItemProvider. Unlike its counterpart, ISettingsMenuProvider, it only provides an item in the settings menu list, as opposed to a button and sub-face. It is implemented in almost exactly the same way--it provides a prefab which the system instantiates and serves back to the implementer. In implementing this menu, I came across the shortcoming that the existing settings menu system assumed a single provider per-type, which required fixes in both serialization and menu implementation.

Resolves #203

One final note: I just tried lowering the minimum scale back down to 0.01. UI raycasts do work sometimes, but everything is unstable in a way that wasn't the case in 5.4. We will have to investigate further if we want to allow editing at extremely small scales.


Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+								if (thumbHeld && otherThumbHeld)
+								{
+									m_AllowScaling = false;
+
+									rayToRay = otherRayOrigin.position - rayOrigin.position;
+									midPoint = rayOrigin.position + rayToRay * 0.5f;
+									var currOffset = midPoint - cameraRig.position;
+
+									cameraRig.position = midPoint - currOffset / currentScale;
+									cameraRig.rotation = Quaternion.AngleAxis(m_StartYaw, Vector3.up);
```
AndrewTHEManeri: Should we be altering rotation at all here if it is just a scale reset?

amirebrahimi:
I'd suggest removing the "possibly fixes" part as that will automatically close that bug. My suggestion is to either verify that it does fix the bug or remove that.

New fly mode is cool!
Also, it's odd to me that the locomotion settings are directly in the Settings menu and not underneath a separate submenu. There's nothing there that designates that it is even for locomotion.

Re: the other changes - while thumbpad press on the Vive probably feels okay, on the Rift it is error prone and easy to lose a multi-selection. Would hope that there could be a better solution that you can come up with @timoni.

Using grip to move can easily trigger scaling. One improvement for this would be to have a small threshold of hand movement where scale doesn't change. Can you try that or discuss with @timoni?

The arc for blink is incorrect at short distances. Would be great if you could fix that in this update, since this is a locomotion update. Otherwise, I can report a separate bug if there isn't one already.

Docs need to be updated

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+			m_MainCamera.nearClipPlane = m_OriginalNearClipPlane;
+			m_MainCamera.farClipPlane = m_OriginalFarClipPlane;
+
+			if (m_ViewerScaleVisuals)
+				ObjectUtils.Destroy(m_ViewerScaleVisuals.gameObject);
+		}
+
+		void CreateViewerScaleVisuals(Transform leftHand, Transform rightHand)
+		{
+			m_ViewerScaleVisuals = ObjectUtils.Instantiate(m_ViewerScaleVisualsPrefab, cameraRig, false).GetComponent<ViewerScaleVisuals>();
```
AndrewTHEManeri: Probably want to cache/show/hide these visuals rather than remaking them all the time

Scripts/Core/EditorVR.Interfaces.cs
```diff
@@ -51,12 +51,22 @@ internal void ConnectInterfaces(object obj, Transform rayOrigin = null)
 					connectInterfaces(obj, rayOrigin);
 			}
 
-			internal void DisconnectInterfaces(object obj)
+			internal void DisconnectInterfaces(object obj, InputDevice device)
```
amirebrahimi: Why this? Looking at how you're calling it you could have just passed the deviceData and since you could have passed that, then why not simply pass the rayOrigin?

Scripts/Core/VRView.cs
```diff
@@ -132,9 +133,8 @@ public void OnEnable()
 			m_Camera.farClipPlane = 1000f;
 
 			// Generally, we want to be at a standing height, so default to that
-			const float kHeadHeight = 1.7f;
 			Vector3 position = m_CameraRig.position;
-			position.y = kHeadHeight;
+			position.y = HeadHeight;
```
amirebrahimi: Is this information provided from the partner SDKs and/or Unity API? Would be good to have the default HeadHeight at 1.7 unless a custom value is provided from the SDK.

Scripts/Interfaces/Entity/IMainMenu.cs
```diff
@@ -23,7 +23,12 @@ public interface IMainMenu : IMenu, IUsesMenuActions, ISelectTool
 		/// <summary>
 		/// The types which provide a settings menu
 		/// </summary>
-		Dictionary<Type, ISettingsMenuProvider> settingsMenuProviders { set; }
+		Dictionary<KeyValuePair<Type, Transform>, ISettingsMenuProvider> settingsMenuProviders { set; }
```
amirebrahimi: Why both menus providers and menu item providers?

Scripts/Modules/SerializedPreferencesModule.cs
```diff
@@ -94,8 +95,10 @@ internal string SerializePreferences()
 				}
 
 				if (item != null)
-					preferences.items.Add(item);
+					payloads[serializer.GetType()] = item;
```
amirebrahimi: Echoing @AndrewTHEManeri's comment, too. Not sure why this is necessary. Was this used for debug purposes at one point? I'd say revert back to what it was.

Tools/LocomotionTool/Scripts/BlinkVisuals.cs
```diff
@@ -180,9 +175,9 @@ void Update()
 				DrawMotionSpheres();
 
 				m_RoomScaleTransform.position = MathUtilsExt.SmoothDamp(m_RoomScaleLazyPosition, m_LocatorRoot.position,
-					ref m_MovementVelocityDelta, 0.2625f, 100f * viewerScale, Time.deltaTime);
+					ref m_MovementVelocityDelta, 0.2625f, 100f * this.GetViewerScale(), Time.deltaTime);
 
-				// Since the room scale visuals are parented under the locator root it is necessary to cache the position each frame before the locator root gets updated
+				// Since the room scale visuals are parented under the locater root it is necessary to cache the position each frame before the locater root gets updated
```
amirebrahimi: Why'd you change the spelling?

mtschoen-unity: Visual Assist was showing it as a typo, which was distracting.  I'm not actually sure if one is more/less correct.

Scripts/Modules/SerializedPreferencesModule.cs
mtschoen-unity: This is there to prevent an error which occurs on deserialization if you do somehow save more than one payload of the same type. I also check against this in the tool, but I thought it was a good idea to leave the safeguard in place.

Should we also print a warning if something tries to serialize the same type twice?

Menus/MainMenu/MainMenu.cs
```diff
@@ -143,77 +158,130 @@ void CreateFaceButtons(List<Type> types)
...
+						if (targetRayOrigin)
+						{
+							this.SelectTool(targetRayOrigin, selectedType);
+							UpdateToolButtons();
+						}
+					});
 				}
-				else if (isSettingsProvider)
+
+				if (isWorkspace)
```
mtschoen-unity: I'm not sure. Generally, a tool which is also a workspace would probably have an issue being either one independently. I actually haven't tested this to see what happens.

I also had a similar issue with the LocomotionTool in that it is a tool and a SettingsMenuItemProvider, but should not actually be included in the tools menu. Hence the added check to see if the tool is in the list of menuTools.

Generally, the reason for this change was because the LocomotionTool is now a type that is two things as far as the menu is concerned. We could alternatively go back to the previous setup and add a new class to delegate the ISettingsMenuItemProvider stuff.

Menus/MainMenu/MainMenu.cs
```diff
@@ -143,77 +158,130 @@ void CreateFaceButtons(List<Type> types)
...
+						var kvp = providerPair.Key;
+						if (kvp.Key == type && (kvp.Value == null || kvp.Value == rayOrigin))
 						{
-							var provider = settingsMenuProviders[selectedType];
-							provider.settingsMenuInstance = m_MainMenuUI.AddSubmenu(buttonData.sectionName, provider.settingsMenuPrefab);
-						});
+							var itemProvider = providerPair.Value;
+							if (buttonData == null)
+							{
+								buttonData = new MainMenuUI.ButtonData
```
mtschoen-unity: Yeah, I went back and forth a lot on how this code was structured. Let's maybe have a chat.

Scripts/Core/VRView.cs
mtschoen-unity: I want to say this used to be done only for Oculus, but I can't quite remember. All I'm doing in this case is exposing the const so I can reset to the same position.

Scripts/Interfaces/ISettingsMenuItemProvider.cs
```diff
@@ -0,0 +1,28 @@
+﻿#if UNITY_EDITOR
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+	/// <summary>
+	/// Decorates types that can provide a sub-menu on the Settings menu. The class should also use a MainMenuItemAttribute
```
mtschoen-unity: I'm not sure what you mean. What should inherit what?

The MainMenuItemAttribute is also problematic if we continue down the road of one class potentially being on the menu twice.

Scripts/Modules/SerializedPreferencesModule.cs
```diff
@@ -94,8 +95,10 @@ internal string SerializePreferences()
...
 			}
+
+			preferences.items.AddRange(payloads.Values);
```
mtschoen-unity: This is again to protect against adding an item of the same type more than once. The dictionary is ensuring that we just overwrite payloads of the same type instead of holding on to them.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+		Preferences m_Preferences;
+
+		ViewerScaleVisuals m_ViewerScaleVisuals;
+
+		GameObject m_BlinkVisualsGO;
+		BlinkVisuals m_BlinkVisuals;
+
+		State m_State = State.Inactive;
+
+		bool m_AllowScaling = true;
```
mtschoen-unity: Yeah. A big improvement in this iteration is splitting out the various types of locomotion into their own method. In fact, the reason this was originally BlinkLocomotionTool is that we had intended for users to explicitly swap different locomotion tools for different types, but then realized that we pretty much want everything on all the time, if possible.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+
+		public GameObject settingsMenuItemPrefab { get { return m_SettingsMenuItemPrefab; } }
+
+		public GameObject settingsMenuItemInstance
+		{
+			set
+			{
+				foreach (var toggle in value.GetComponentsInChildren<Toggle>())
+				{
+					if (toggle.isOn)
```
mtschoen-unity: Yeah. It's basically a way to avoid having to create a new script to put on the prefab just for the purpose of labeling the toggles. I'm happy to do it that way, if you prefer, or come up with a better heuristic (say, first child or something)

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+					{
+						m_FlyToggle = toggle;
+						toggle.onValueChanged.AddListener(isOn =>
```
mtschoen-unity: Nah, one-time. This happens once after you instantiate the prefab, and when you disable the tool (or in this case when you close EVR) we destroy the menu object

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+						{
+							if (m_BlockValueChangedListener)
+								return;
+
+							// m_Preferences on all instances refer
+							m_Preferences.blinkMode = !isOn;
+							foreach (LocomotionTool linkedObject in linkedObjects)
```
mtschoen-unity: Hm. I actually switched to this approach just to cut out that boilerplate line. Either way we need to use the class name, so a refactor would have to pick it up. My thinking going forward is only to use var if we don't need the cast. I'm open to discuss this, however.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+							{
+								if (linkedObject != this)
+								{
+									linkedObject.m_BlockValueChangedListener = true;
+									//linkedObject.m_ToggleGroup.NotifyToggleOn(isOn ? m_FlyToggle : m_BlinkToggle);
+									// HACK: Toggle Group claims these toggles are not a part of the group
```
mtschoen-unity: Will do. I need to investigate a little further and/or come up with a simpler repro case.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+			m_BlinkVisuals.enabled = false;
+			m_BlinkVisualsGO.transform.parent = rayOrigin;
+			m_BlinkVisualsGO.transform.localPosition = Vector3.zero;
+			m_BlinkVisualsGO.transform.localRotation = Quaternion.identity;
+
+			m_MainCamera = CameraUtils.GetMainCamera();
+			m_OriginalNearClipPlane = m_MainCamera.nearClipPlane;
+			m_OriginalFarClipPlane = m_MainCamera.farClipPlane;
+
+			Shader.SetGlobalFloat(k_WorldScaleProperty, 1);
```
mtschoen-unity: Hm... The interface is just about accessing the scale property on the camera rig transform. This is the place where we actually do the scaling, so any other place would have to poll the value in an update loop or something. Can you explain more about how you are thinking of doing this?

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+
+				if (locomotionTool.m_State != State.Inactive)
+					return;
+			}
+
+			DoTwoHandedScaling(consumeControl);
+
+			if (!m_Scaling)
+			{
+				if (!m_WasRotating)
```
mtschoen-unity: I'll give this a try

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,559 @@
...
+									var yawSign = Mathf.Sign(Vector3.Dot(Quaternion.AngleAxis(90, Vector3.down) * m_StartDirection, rayToRay));
+									var currentYaw = m_StartYaw + Vector3.Angle(m_StartDirection, rayToRay) * yawSign;
+									var currentRotation = Quaternion.AngleAxis(currentYaw, Vector3.up);
+									midPoint = currentRotation * midPoint * currentScale;
+
+									cameraRig.position = m_StartPosition + m_StartMidPoint - midPoint;
+									cameraRig.localScale = Vector3.one * currentScale;
+									cameraRig.rotation = currentRotation;
+
+									m_MainCamera.nearClipPlane = m_OriginalNearClipPlane * currentScale;
```
mtschoen-unity: Hm. I definitely saw issues with this when scaling myself down very small. Let me test on my end

Tools/LocomotionTool/LocomotionTool.cs
mtschoen-unity: Sounds good.

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -114,10 +118,20 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 				}
 			}
 
+			var selectionInput = (SelectionInput)input;
```
mtschoen-unity: Good catch. Simple oversight

mtschoen-unity:
Thanks for the reviews. I'm going to go and make the obvious changes. It sounds like you both had issues with Thumbstick for multi-select. I'll be honest: I don't love it either. Adding the Activation Threshold for the radial menu definitely helped, but it seems like that's not enough.

@amirebrahimi, you make a good point about the settings being in-line in the menu. I still think it is valuable to have the option to add a small item to the settings menu, rather than needing an entire sub-face for everything. I think a good compromise would be to add a title "Locomotion Mode:" to the prefab, and maybe reducing the height of the toggle button. I'm also not opposed to deciding not to implement the SettingsMenuItem feature if we think it's a bad design. Thoughts, @timoni?

I've been racking my brain about a better way to avoid overlaps between these "always-on" types of interactions. We had also tried options where grip is still multi-select, but will block crawl locomotion if an object is selected, or if you are currently hovering an object, but it was hard to understand why crawl wasn't working.

Another option is to further break down the locomotion modes so that crawl is also bound to the menu/B button, but I think it feels much better using the grips. I'm not sure how best to go forward. In my opinion, this is the best compromise, but perhaps there is a better way.

Menus/MainMenu/MainMenu.cs
AndrewTHEManeri: discussion result: Leaving this one alone

Tools/LocomotionTool/LocomotionTool.cs
mtschoen-unity: The idea is to undo what you just did with the two-handed scale, since you might just be wanting to reset your scale. @timoni, thoughts?

Menus/MainMenu/MainMenu.cs
AndrewTHEManeri: discussion result : doing name sanitizing.  If this ends up over-aggressive at some point we'll introduce a way to specify custom clean names

Scripts/Modules/SerializedPreferencesModule.cs
amirebrahimi: Hmm...Useful for us internally, but would be confusing to someone else. We would need some way to turn it on only for our builds. You could look for an internal define that only shows up when building from trunk?

AndrewTHEManeri: Discussion result: Taking this a step further actually.  Storing the prefs in a dictionary and converting to list to serialize/deserialize.  Also spitting out warning if double-serialization tries to happen.

Tools/LocomotionTool/LocomotionTool.cs
AndrewTHEManeri: Discussion result: This is only happening here.  If/when that changes we'll centralize these things to ISetViewerScale

Scripts/Core/VRView.cs
mtschoen-unity: We should also check if the tracking is configured for sitting or standing and not apply the offset if it is standing/roomscale. Since I am just exposing this for the reset feature, I would like to punt this to a later improvement. I can create an issue if needed.

Scripts/Core/EditorVR.Interfaces.cs
mtschoen-unity: This was just to stay consistent with the way ConnectInterfaces is called.

Scripts/Interfaces/Entity/IMainMenu.cs
mtschoen-unity: Menu Item provider is for adding single items like this toggle group to the settings menu, as opposed to entire sub-faces

Tools/LocomotionTool/LocomotionTool.cs
mtschoen-unity: Yeah. If you scale yourself down, the near clip plane starts clipping out the EVR objects.

Scripts/Core/VRView.cs
AndrewTHEManeri: Yeah, let's issue this so we don't lose it

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -30,7 +30,8 @@ internal enum MenuHideFlags
 				NearWorkspace = 1 << 2,
 			}
 
-			readonly Dictionary<Type, ISettingsMenuProvider> m_SettingsMenuProviders = new Dictionary<Type, ISettingsMenuProvider>();
+			readonly Dictionary<KeyValuePair<Type, Transform>, ISettingsMenuProvider> m_SettingsMenuProviders = new Dictionary<KeyValuePair<Type, Transform>, ISettingsMenuProvider>();
```
AndrewTHEManeri: We're going to punt on this for now - we haven't found a solution that was convincingly more elegant than this

AndrewTHEManeri:
There are some open issues here in regards to the actual UI menu and other visual flare type-things - but I feel they are more appropriate for a subsequent PR rather than this one - and these new locomotion modes feel really good.  I'll be adding issues for them if they do not already exist.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -0,0 +1,599 @@
...
+
+				consumeControl(m_LocomotionInput.forward);
+				return true;
+			}
+
+			m_Rotating = false;
+			return false;
+		}
+
+		bool DoCrawl(ConsumeControlDelegate consumeControl)
```
dunity: Doesn't look like we're using this return value.  Though it does match the "Do" functions.

dunity: After speaking with @Schoen , me mentioned that this is intentional.

Scripts/Core/EditorVR.Interfaces.cs
amirebrahimi: I don't think it is needed.

amirebrahimi: @AndrewTHEManeri - if you're going to fix, then take a look at my original comment.

AndrewTHEManeri:
DisconnectInterfaces altered back - If we wish to maintain parity between the implementations of Connect and Disconnect, another option to make an Extension method for InputDevice to return a RayOrigin - at which point the calls to Connect to reference the RayOrigin version exclusively as well.

AndrewTHEManeri:
Manual merge test was successful, performing auto-merge now

PR #226, opened 2017-06-21
Haptics bugfixes

mtschoen-unity:
A couple things snuck through in the last PR.
- Workspaces with zoom sliders didn't load correctly because of a null ref in WorkspaceButton
- IRayToNode was using old-style functionality injection
- Project workspace lists were missing their haptic pulse object

### DIFF
--- a/Scripts/ListView/ListViewControllerBase.cs
+++ b/Scripts/ListView/ListViewControllerBase.cs
@@ -87,11 +87,7 @@ namespace ListView
 			UpdateView();
 
 			if (m_Scrolling)
-			{
-				if (!m_ScrollPulse)
-				Debug.Log(this, this);
 				this.Pulse(null, m_ScrollPulse);
-			}
 		}
 
 		protected abstract void Setup();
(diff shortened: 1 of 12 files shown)


Scripts/ListView/ListViewControllerBase.cs
```diff
@@ -87,7 +87,11 @@ void Update()
 			UpdateView();
 
 			if (m_Scrolling)
+			{
+				if (!m_ScrollPulse)
```
AndrewTHEManeri: Probably don't want this logging here

Workspaces/Common/Scripts/WorkspaceButton.cs
```diff
@@ -211,7 +208,8 @@ void Awake()
 			if (m_SwapIconsOnClick && m_AlternateIconSprite)
```
AndrewTHEManeri: the if(m_Button) check needs to surround this whole block or the next line will have a nullref

Scripts/ListView/ListViewControllerBase.cs
mtschoen-unity: Ah yep. My bad there.

PR #235, opened 2017-07-04
Fix Show EditorVR Objects bug (#223)

### LINKED ISSUE #223: "Show EditorVR Objects" change doesn't take effect until restarting Unity
One would expect this change to take effect next time you started EditorVR, but you must restart the entire editor for the change to take effect.

mtschoen-unity:
Got bored on the plane from Iceland and decided to take a crack at this one. The setting only took effect if you recompiled because we set the hideflags value in a static constructor. It did not appear to be needed before Awake, so I put it there.

Fixes #223

### DIFF
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -438,6 +438,8 @@ namespace UnityEditor.Experimental.EditorVR.Core
 
 		static EditorVR()
 		{
+			ObjectUtils.hideFlags = defaultHideFlags;
+
 			if (!PlayerSettings.virtualRealitySupported)
 				Debug.Log("<color=orange>EditorVR requires VR support. Please check Virtual Reality Supported in Edit->Project Settings->Player->Other Settings</color>");
 
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -96,6 +96,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
 
 		void Awake()
 		{
+			ObjectUtils.hideFlags = defaultHideFlags;
 			Nested.evr = this; // Set this once for the convenience of all nested classes 
 			m_DefaultTools = defaultTools;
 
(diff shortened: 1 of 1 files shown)


Scripts/Core/EditorVR.cs
```diff
@@ -96,6 +96,7 @@ static void ResetPreferences()
 
 		void Awake()
 		{
+			ObjectUtils.hideFlags = defaultHideFlags;
```
amirebrahimi: I think you might break existing functionality this way, but haven't tested. By moving this to awake, the default hide flags would not be set in time for the editing context manager on initialization. So, you'll need to test whether things are hidden correctly on first start of the an editorvr installation. My guess is no.

Also, IIRC static constructors should be called even on re-compile, so it's odd that this is not getting set.

AndrewTHEManeri: He's saying it _already_ works on re-compile.  What is missing is that restarting EditorVR (closing and re-opening the window) wouldn't get the updated hide-flags.

mtschoen-unity: Exactly. One would expect this to take effect the next time I restart EVR (or immediately, while EVR is running, but that would be going above and beyond).

I had considered leaving the one in the static constructor as well, but that seemed redundant.

I put break points on all the places where the value is used, and none are used before Awake. This may not always be the case, of course. Are we in favor of putting it back into the static ctor as well as leaving it here?

Also, @amirebrahimi would you like to continue to be added to PRs on core changes?

amirebrahimi: We talked about this live, but I'll copy here for @AndrewTHEManeri - both locations will keep the set of ObjectUtils.hideFlags (i.e. static constructor and Awake) and Andrew makes the call for PRs. I may add comments here and there, but won't create a review unless explicitly asked.

mtschoen-unity:
Good catch. Not sure how I didn't see it. I'll find another way.

PR #238, opened 2017-07-06
Fix errors and warnings in HapticsModule caused by not having Oculus Utilities

mtschoen-unity:
First fix is to wrap LateUpdate in #if ENABLE_OVR_INPUT to remove call to missing OVRHaptics class;
Second fix is to leave the usage of m_MasterIntensity in Pulse in order to eliminate assigned but never used warning.


AndrewTHEManeri:
Manual merge successful, merging through github

PR #240, opened 2017-07-10
Fix Clone/Cut/Copy/Paste so that new objects maintain prefab links

### LINKED ISSUE #38: Duplicated prefab is not a prefab
In the Unity Editor if I copy-and-paste (or duplicate) a prefab in my scene, the new copy is also a prefab.  But in EditorVR if I select a prefab then use copy-and-paste or duplicate via the radial menu, then the new copy is an hierarchy of game objects rather than a prefab.  

In the attached screen shot the Hierarchy shows several Tank prefabs.  In EditorVR I selected one of these and used the Duplicate button via radial menu.  Then I selected the prefab again and pressed Copy on the radial menu, following by Paste.  These operations generated Tank(Clone) and Tank(Clone)(Clone) which are not prefabs.  Rather, they are hierarchies of primitive objects.  (Prefabs are blue in Hierarchy.) 

tanks

mtschoen-unity:
### Purpose of this PR
The current Clone/Cut/Copy/Paste actions create new objects which are not properly lined to prefabs like the originals.

### Testing status
Tested on a few sample prefabs, including the ArcheryWeeble from our SIGGRAPH talk (basically the same as the one in SteamVR). Easy way to know it's not working is if you accidentally "pick apart" a cloned object whose root transform does not have a renderer.

### Technical risk
Low; Only affects these 4 actions and uses an identical method to normal editor actions.

### Comments to reviewers
I'm not thrilled about using stuff in Unsupported, but this was the only way I could find to properly duplicate prefabs. This is what happens when you press ctrl-D, ctrl-C, ctrl-V, and ctrl-X. I tried using PrefabUtility to do the instantiate, or hook up the new object to the old prefab. If you instantiate the prefab, you then have to go make all the override changes, and if you try to call connectprefab, you get an error about hideflags.

Fixes #38

### DIFF
--- a/Actions/Clone.cs
+++ b/Actions/Clone.cs
@@ -1,5 +1,4 @@
 ﻿#if UNITY_EDITOR
-using System;
 using UnityEditor.Experimental.EditorVR.Utilities;
 using UnityEngine;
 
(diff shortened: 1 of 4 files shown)


AndrewTHEManeri:
Manual Merge Successful, merging through github now

PR #241, opened 2017-07-10
Smooth Camera Locomotion

mtschoen-unity:
### Purpose of this PR
The current SmoothCamera will also smooth out movements due to locomotion, which looks awkward (controllers fly around in preview window). This PR eliminates this issue

### Testing status
Tested on current development with Preview Camera enabled. Blink, fly, crawl, and world scale do not create the undesired effect.

### Technical risk
Low; Only affects Preview Camera

### Comments to reviewers
There's no issue for this, but it was bugging me :)

### DIFF
--- a/Scripts/Helpers/VRSmoothCamera.cs
+++ b/Scripts/Helpers/VRSmoothCamera.cs
@@ -9,7 +9,7 @@ namespace UnityEditor.Experimental.EditorVR.Helpers
 	/// </summary>
 	[RequireComponent(typeof(Camera))]
 	[RequiresLayer(k_HMDOnlyLayer)]
-	sealed class VRSmoothCamera : MonoBehaviour, IPreviewCamera, IUsesViewerScale
+	sealed class VRSmoothCamera : MonoBehaviour, IPreviewCamera
 	{
 		/// <summary>
 		/// The camera drawing the preview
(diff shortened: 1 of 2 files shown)


Scripts/Helpers/VRSmoothCamera.cs
```diff
@@ -78,12 +83,11 @@ void LateUpdate()
 			m_SmoothCamera.stereoTargetEye = StereoTargetEyeMask.None;
 			m_SmoothCamera.fieldOfView = m_FieldOfView;
 
-			m_Position = Vector3.Lerp(m_Position, m_VRCamera.transform.position, Time.deltaTime * m_SmoothingMultiplier);
-			m_Forward = Vector3.Lerp(m_Forward, m_VRCamera.transform.forward, Time.deltaTime * m_SmoothingMultiplier);
+			m_Position = Vector3.Lerp(m_Position, m_VRCamera.transform.localPosition, Time.deltaTime * m_SmoothingMultiplier);
+			m_Forward = Vector3.Lerp(m_Forward, m_VRCamera.transform.localRotation * Vector3.forward, Time.deltaTime * m_SmoothingMultiplier);
```
AndrewTHEManeri: You could potentially get an error in Quaternion.LookRotation if the start and end forward vectors ever cancel out to zero during the lerp.
Any reason we can't slerp the local rotations instead?

mtschoen-unity: I thought the LookRotation error was only for input vectors that are too close to zero magnitude. Is there a good way of avoiding this other edge case? Just check if the dot product is too close to 1 or the distance is too close to 0?

The reason we're using a forward direction and not the whole rotation is to eliminate roll. AFAIK, trying to pull roll out of a quaternion doesn't work in the general case without gimbal lock somewhere. I didn't write the original code, though, so I can't really say.

AndrewTHEManeri:
Manual merge successful, performing merge through github

PR #243, opened 2017-07-10
Transform improvements

### LINKED ISSUE #232: Transform manipulator gizmo pieces should highlight when selected
I would like the transform manipulator gizmo to show a bit more obviously that one part or another is being selected by the controller and will take effect.  I notice this mostly with the planar controls as the other gizmos have some kind of feedback, but I the highlight should work well on all of them.

### LINKED ISSUE #233: Object's selection state will now update until the object is deselected
A few examples
  Direct selecting an object (in the miniworld or main world) and then ray selecting it will prevent the transform gizmo from appearing.  It will not appear until it has been deselected, even though haptics are telling you you have a done a 'new' selection.

This also means that if you select an object with one hand, the radial menu will not switch hands even if you reselect it with the other - until it has been deselected first.

mtschoen-unity:
### Purpose of this PR
Make it easier to select handles on the manipulator (and fix a bug). Adds a "scale bump" to linear and plane handles.

### Testing status
It works

### Technical risk
Low; Only affects manipulator

### Comments to reviewers
The bug fix is for a null reference which can occur when there are selected transforms but no active transform. I can't remember exactly how I got into this state (maybe a group selection in the heirarchy, or lasso in the scene) but I got into this state while working and added a fix.

Otherwise, the scale bump thing was already happening on the sphere handle. I just applied it to the other handles. I couldn't do it on the rotation handles because scaling them moves the raycast target and can "flicker" because it moves on hover. Plus, it just doesn't look quite right. We could add a "fatter" torus model if we really want to.

ScaleBump is a setting because we use the handle scripts elsewhere where we don't want the scale bump.

Bonus! Fixes #232, Fixes #233


mtschoen-unity:
@AndrewTHEManeri at long last, this one's wrapped up. The last three commits represent some pretty significant changes, so be sure to give it a thorough review.

Manipulators/BaseManipulator.cs
```diff
@@ -87,24 +122,185 @@ void OnCameraPreRender(Camera camera)
...
 			cameraPosition = worldToCameraMatrix.MultiplyPoint3x4(cameraPosition);
 			var delta = worldToCameraMatrix.inverse.MultiplyPoint3x4(cameraPosition - manipulatorPosition) - originalCameraPosition;
 			transform.localScale = Vector3.one * delta.magnitude * k_BaseManipulatorSize;
 		}
 
-		protected void OnDragStarted()
+		protected virtual void UpdateHandleTip(BaseHandle handle, HandleEventData eventData, bool active)
+		{
+			var rayOrigin = eventData.rayOrigin;
+			HandleTip handleTip;
```
AndrewTHEManeri: there are a lot of handletips here, some that are handletips, and some that are named handletip but are renderers

Scripts/Interfaces/Capability/IManipulatorController.cs
```diff
@@ -0,0 +1,20 @@
...
+	{
+		/// <summary>
+		/// Show or hide the manipulator(s)
+		/// </summary>
+		bool manipulatorVisible { set; }
+
+		/// <summary>
+		/// Whether the manipulator(s) are in a drag state
+		/// </summary>
+		bool manipulatorDragging { get; }
```
mtschoen-unity: This is what is used by IGetManipulatorDragState. We loop through the ManipulatorControllers and check manipulatorDragging

Manipulators/StandardManipulator.cs
```diff
@@ -52,40 +36,68 @@ void Update()
...
+
+		protected override void OnHandleDragEnded(BaseHandle handle, HandleEventData eventData)
+		{
+			base.OnHandleDragEnded(handle, eventData);
+
+			if (handle.IndexOfDragSource(eventData.rayOrigin) > 0)
+				return;
 
-			OnDragEnded(eventData.rayOrigin);
+			if (handle is SphereHandle)
```
mtschoen-unity: I see this as part of the StandardManipulator. The SphereHandle and RadialHandle provide a way of processing drag events along the surface of a sphere, or into rotation deltas. As with the rest of this PR, I tried to keep any visual behavior out of the Handle classes themselves.

It helps to remember that the sphere object that renders the object with the spherehandle on it has nothing to do with the "sphere" in spherehandle. That sphere is the sphere that encompasses the controller and defines the surface along which you will drag the object.

Scripts/Handles/BaseHandle.cs
```diff
@@ -77,23 +90,32 @@ protected virtual HandleEventData GetHandleEventData(RayEventData eventData)
...
+			var rayOrigin = eventData.rayOrigin;
+			m_DragSources.Add(rayOrigin);
+			startDragPositions[rayOrigin] = eventData.pointerCurrentRaycast.worldPosition;
 
 			var handleEventData = GetHandleEventData(eventData);
 
 			//Double-click logic
-			var timeSinceLastClick = (float) (DateTime.Now - m_LastClickTime).TotalSeconds;
+			var timeSinceLastClick = (float)(DateTime.Now - m_LastClickTime).TotalSeconds;
 			m_LastClickTime = DateTime.Now;
```
mtschoen-unity: Hm, I suppose you're right. Seems like a pretty slim chance of happening, but we might as well be consistent.

Now that we're on the subject, we should remember to check the double click behavior when moving to the new input system. Currently, it is possible to double click so fast that we dont catch the return to a zero-value, and essentially you can double click between frames. If we send input events more quickly than the update rate, we can probably fix this issue.

AndrewTHEManeri:

PR #245, opened 2017-07-10
Skinned Mesh Support

mtschoen-unity:
### Purpose of this PR
Skinned meshes can't be selected or direct manipulated because they weren't added to the spatial hash.

### Testing status
Tested with various characters from Polygon Adventure pack and bow from SteamVR

### Technical risk
Medium; Might see some errors trying to use certain skinned meshes in MeshColliders.

### Comments to reviewers
I can't exactly remember why we didn't add them originally. I think we saw some errors on certain models, but I don't see anything now. As you might expect, once the character has been posed, you can't hover-and-select all visible parts of the model, and you see an outline of the T-pose on hover. This is an inherent limitation of the MeshCollider component, which can't deform skinned meshes (as far as I know). We could try bringing back the PixelRaycastModule which did regular Scene Picking using a GPU readback, but removing it was a nice boost to performance, so I'd rather not.

### DIFF
--- a/Scripts/Helpers/VRSmoothCamera.cs
+++ b/Scripts/Helpers/VRSmoothCamera.cs
@@ -73,7 +73,7 @@ namespace UnityEditor.Experimental.EditorVR.Helpers
 			m_SmoothCamera.targetTexture = m_RenderTexture;
 			m_SmoothCamera.targetDisplay = m_TargetDisplay;
 			m_SmoothCamera.cameraType = CameraType.Game;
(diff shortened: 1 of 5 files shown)


mtschoen-unity:
Huh. I stand corrected. I did not know about BakeMesh (nor did I know we've had it since 4.0!) I'll get this in there

mtschoen-unity:
Again, thanks for pushing me just a little farther 👍

AndrewTHEManeri:
It even works if you keep it highlighted with the other hand and move bones with the first.
Manual merge successful, performing github merge.

PR #248, opened 2017-07-11
Hierarchy prefab colors

mtschoen-unity:
### Purpose of this PR
Set text color in hierarchy to blue for prefabs. Also adds a selected-and-hovered state because otherwise there was no hover for the selected object.

### Testing status
Tested in Polygon Adventure scene with a few prefabs in the root

### Technical risk
None

### Comments to reviewers
Forgot to get to this one before lunch. I had to reduce the brightness of the hover color slightly to make the blue text readable on hover.


AndrewTHEManeri:

PR #251, opened 2017-07-14
World Scale Performance Fixes

### LINKED ISSUE #237: CollectHierarchyData is slow for large scenes with deep hierarchies
I noticed this when trying to world-scale in the Synty Simple Airport scene. When the visuals appear and disappear, we get an OnHierarchyChanged event, which takes ~500ms to re-generate the hierarchy data.

The profiler shows some GC allocation, but the biggest culprit is the HierarchyProperty.Previous() method, which takes ~30ms per call. The reason for this is that it doesn't actually "roll back" the iterator, but resets it and iterates back through the whole tree.

The fix will likely be to refactor CollectHierarchyData into a while loop instead of a recursive call.

mtschoen-unity:
### Purpose of this PR
Improve the performance of hierarchy updates and intersection calls that were tanking performance while doing two-handed world scaling.

### Testing status
The issue was discovered in the Synty Airport Demo scene http://u3d.as/gpK

On Development the call to HierarchyChanged when you do world scaling takes 463ms, and on this branch it takes 110ms. This is still not ideal, so we might have to spread this out over a few frames like how the Project data is created. For now, however, I don't really notice the slowdown since this only happens once when the hierarchy is changed. The problem with doing the setup as a coroutine is that OnHierarchyChanged happens quite frequently, so we will need a way to either cancel the current, out-of-date process, or queue up change requests because they will likely occur during an existing traversal of the hierarchy. This might also not work if HierarchyProperty doesn't play nice with scene changes in the middle of a traversal.

Secondly, one of the biggest issues this change addresses is that, while scaling the world, the intersection module does a lot (~120/frame) of tests, presumably because the cone rays are all compressed down into a single point when they are hidden. The fix was simply to disable the tester when DefaultProxyRay hides the cone.

Between these two fixes (and a few other places where I got rid of GC allocation), performance in the Airport Demo scene is much, much better.

### Technical risk
Medium; This is a total rewrite of how we turn the HierarchyProperty into data for our workspace, so there may well be edge cases that we haven't tested for. I was quite thorough in my own testing while working on this and I think I have accounted for most types of hierarchies.

### Comments to reviewers
This turned into a bit of a rabbit hole, but it's a major improvement. That Airport scene isn't even that dense, but performance was awful. The code itself for parsing the hierarchy is a little cleaner and easier to understand as long as you are familiar with using a stack to unwind a recursive method.

I'm hoping this will also greatly improve performance with SceneFusion, but that's just a hunch.

Fixes #237

### DIFF
--- a/Scripts/Modules/HierarchyModule.cs
+++ b/Scripts/Modules/HierarchyModule.cs
@@ -17,7 +17,8 @@ namespace UnityEditor.Experimental.EditorVR.Modules
 		readonly List<GameObject> m_IgnoreList = new List<GameObject>();
 
 		// Local method use only -- created here to reduce garbage collection
-		readonly HashSet<string> m_CompareTypes = new HashSet<string>();
+		readonly Stack<HierarchyData> m_DataStack = new Stack<HierarchyData>();
+		readonly Stack<int> m_SiblingIndexStack = new Stack<int>();
 
 		void Awake()
 		{
@@ -92,10 +93,10 @@ namespace UnityEditor.Experimental.EditorVR.Modules
 
 			var hasChanged = false;
 			var lastDepth = 0;
-			var dataStack = new Stack<HierarchyData>();
-			var siblingIndexStack = new Stack<int>();
-			dataStack.Push(null);
-			siblingIndexStack.Push(0);
+			m_DataStack.Clear();
+			m_SiblingIndexStack.Clear();
+			m_DataStack.Push(null);
+			m_SiblingIndexStack.Push(0);
 			while (m_HierarchyProperty.Next(null))
 			{
 				var instanceID = m_HierarchyProperty.instanceID;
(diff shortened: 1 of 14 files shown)


Scripts/Modules/HierarchyModule.cs
```diff
@@ -60,141 +79,143 @@ List<string> GetFilterList()
 
 		List<HierarchyData> GetHierarchyData()
 		{
-			if (m_HierarchyData == null)
-				return new List<HierarchyData>();
-
-			return m_HierarchyData.children;
+			return m_HierarchyData ?? new List<HierarchyData>();
```
AndrewTHEManeri: Any reason to not assign m_HierarchyData to a new List<HierarchyData>() instead of just returning it?

AndrewTHEManeri:
Manual merge successful, performing github merge now

Scripts/Modules/HierarchyModule.cs
mtschoen-unity: Actually, we initialize it by default so it should never be null. We also try to re-use the same list as much as possible, so this could actually just turn into return m_HierarchyData;

Workspaces/HierarchyWorkspace/Scripts/HierarchyListViewController.cs
```diff
@@ -41,6 +41,38 @@ sealed class HierarchyListViewController : NestedListViewController<HierarchyDat
...
+		public override List<HierarchyData> data
+		{
+			set
+			{
+				base.data = value;
+
+				if (m_Data != null && m_Data.Count > 0)
+				{
+					// Remove any objects that don't exist any more
+					var missingKeys = m_Data.Select(d => d.index).Except(m_ExpandStates.Keys);
```
mtschoen-unity: Hm. Fair point. I've been meaning to do another GC pass after this landed, so I'll see if this shows up on the profiler.

PR #252, opened 2017-07-15
Transform changes hotfix

mtschoen-unity:
### Purpose of this PR
Fix a bug introduced in review on #243.  I was going over the changes with @dunity and noticed that the handeltip on radial handles was rotating in place in the middle of the circle, rather than at the radius. While I was at it, I also incorporated some of his feedback, namely to make the handles more opaque on hover and increase the scale bump on linear and planar handles.

I also noticed that there was a bug with #235. I didn't actually set the hideflags value on ObjectUtils.hideFlags, so no new objects got the proper hideflags 🤦‍♂️ 

### Testing status
Tested more thoroughly this time :)

### Technical risk
Low: Just a fix to manipulators

### Comments to reviewers
The actual fix is just 0cbbb6e. The rest is design feedback from @dunity.

We wanted to make the hover state go fully opaque, but couldn't think of a good way to let the handle tip (arrow/cone) intersect with the handle properly while also drawing over the scene object. In other words, since the StandardOverlay shader doesn't test depth, opaque objects don't overlap properly. If we do test depth, we don't draw on top of scene objects.


AndrewTHEManeri:

PR #255, opened 2017-07-18
add unit tests for common extension methods & utilities

stella3d:
the beginning of real unit tests for EditorVR

I chose to write tests for the commonly used utilities & extension methods first, not because they're the most likely to break or the most important code (they aren't), but because other code relies on them & their clear definition & small scope are ideal for establishing nice testing patterns.

These are meant to be thorough but not exhaustive - i searched for the number of references to each function in the project & made sure i only wrote tests for things we're using.

I'm already working on the next batch of tests, which will be around more "core" stuff like Contexts, Modules, & initialization.

### DIFF
--- a/Tests/Editor/Unit/Extension Methods/TransformExtensionsTests.cs
+++ b/Tests/Editor/Unit/Extension Methods/TransformExtensionsTests.cs
@@ -1,0 +1,31 @@
+﻿using NUnit.Framework;
+using UnityEngine;
+using UnityEditor.Experimental.EditorVR.Extensions;
+
+namespace UnityEditor.Experimental.EditorVR.Tests.Extensions
+{
+    [TestFixture]
+    public class TransformExtensionsTests
+    {
+        GameObject go;
+
+        [OneTimeSetUp]
+        public void Setup()
+        {
+            go = new GameObject();
+        }
+
+        [Test]
+        public void TransformBounds_TranslatesLocalBoundsToWorld()
+        {
+            var size = new Vector3(2, 2, 2);
+            var offset = new Vector3(1, 2, 3);
+            go.transform.position += offset;
+            var localBounds = new Bounds(go.transform.position, size);
+            var worldBounds = go.transform.TransformBounds(localBounds);
+
+            Assert.AreEqual(localBounds.center, worldBounds.center - offset);
+            Assert.AreEqual(localBounds.extents, worldBounds.extents);
+        }
+    }
+}
--- a/Tests/Editor/Unit/Extension Methods/TransformExtensionsTests.cs
+++ b/Tests/Editor/Unit/Extension Methods/TransformExtensionsTests.cs
@@ -8,6 +8,7 @@ namespace UnityEditor.Experimental.EditorVR.Tests.Extensions
     public class TransformExtensionsTests
     {
         GameObject go;
+        float delta = 0.00000001f;
 
         [OneTimeSetUp]
         public void Setup()
@@ -21,11 +22,11 @@ namespace UnityEditor.Experimental.EditorVR.Tests.Extensions
             var size = new Vector3(2, 2, 2);
             var offset = new Vector3(1, 2, 3);
             go.transform.position += offset;
-            var localBounds = new Bounds(go.transform.position, size);
-            var worldBounds = go.transform.TransformBounds(localBounds);
+            var local = new Bounds(go.transform.position, size);
+            var world = go.transform.TransformBounds(local);
 
-            Assert.AreEqual(localBounds.center, worldBounds.center - offset);
-            Assert.AreEqual(localBounds.extents, worldBounds.extents);
+            Assert.That(local.center, Is.EqualTo(world.center - offset).Within(delta));
+            Assert.That(local.extents, Is.EqualTo(world.extents).Within(delta));
         }
     }
 }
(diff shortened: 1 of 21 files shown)


Tests/Editor/Unit/Extension Methods/TransformExtensionsTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        [Test]
+        public void TransformBounds_TranslatesLocalBoundsToWorld()
+        {
+            var size = new Vector3(2, 2, 2);
+            var offset = new Vector3(1, 2, 3);
+            go.transform.position += offset;
+            var localBounds = new Bounds(go.transform.position, size);
+            var worldBounds = go.transform.TransformBounds(localBounds);
+
+            Assert.AreEqual(localBounds.center, worldBounds.center - offset);
```
AndrewTHEManeri: For these asserts specifically we might want to test to a really small delta instead of an exact equality.  We are transforming vectors through bounds and while it is working now it is just a little more future proof.

Tests/Editor/Unit/Extension Methods/Vec3ExtensionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            var maxY = new Vector3(0f, 2f, 1f);
+            Assert.AreEqual(maxY.MaxComponent(), maxY.y);
+            var maxZ = new Vector3(0f, 1f, 2f);
+            Assert.AreEqual(maxZ.MaxComponent(), maxZ.z);
+        }
+
+        [Test]
+        public void AveragedComponents()
+        {
+            var vec3 = new Vector3(4f, 2f, 6f);
```
AndrewTHEManeri: For this test also test to an accuracy level instead of exact

Tests/Editor/Unit/Extension Methods/Vec3ExtensionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            Assert.AreEqual(vec3.AveragedComponents(), 4f);
+            vec3 = new Vector3(-4f, 0f, 1f);
+            Assert.AreEqual(vec3.AveragedComponents(), -1f);
+        }
+
+        [Test]
+        public void Inverse_PositiveValues()
+        {
+            var vec3 = new Vector3(2f, 4f, 10f);
+            Assert.AreEqual(new Vector3(.5f, .25f, .1f), vec3.Inverse());
```
AndrewTHEManeri: Again here

Tests/Editor/Unit/Extension Methods/Vec3ExtensionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [Test]
+        public void Inverse_NegativeValues()
+        {
+            var vec3 = new Vector3(-10f, -4f, -2f);
+            Assert.AreEqual(new Vector3(-.1f, -.25f, -.5f), vec3.Inverse());
```
AndrewTHEManeri: And here

Tests/Editor/Unit/Utilities/ObjectUtilsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            var gameObjectA = new GameObject();
+            gameObjectA.transform.position += new Vector3(-5, -2, 8);
+            var gameObjectB = new GameObject();
+            gameObjectB.transform.position += new Vector3(2, 6, 4);
+            var transforms = new Transform[] { gameObjectA.transform, gameObjectB.transform };
+
+            var bounds = ObjectUtils.GetBounds(transforms);
+            Bounds expected = new Bounds(new Vector3(-1.5f, 2f, 6f), new Vector3(7f, 8f, 4f));
+            Assert.AreEqual(expected, bounds);
```
AndrewTHEManeri: Use a delta here

Tests/Test Helpers/TestImage.cs
```diff
@@ -0,0 +1,13 @@
+﻿using UnityEngine.UI;
+
+// if you create a custom class in a file that doesn't match the class name,
+// after the tests run Unity will complain that the compiler that imported 
+// "TestImage" isn't available anymore, so they go in their own files
+namespace UnityEditor.Experimental.EditorVR.Tests.TestHelpers
+{
+    public class TestImage : Graphic
```
mtschoen-unity: Confused about this class. Doesn't this follow the rules?

stella3d: which rules ?  basically, i just needed a `Graphic` of some sort to test with inside `GetMaterialClone_ClonesGraphicMaterial` , but you have to drive from `Graphic`, can't instantiate one directly, and wanted a class i knew wouldn't change outside of the test code.

mtschoen-unity: Ah, so maybe the comment is just out of date? It seems to be talking about having a class name that is different from the filename. At the very least, it doesn't seem to say the same thing you just wrote. I'd just get rid of that comment, or replace it with the thing you just replied with.

I had assumed because of the comment that this was trying to test a case where the class name is different from the filename.

stella3d: i can just remove the comment - the class exists in a separate file just to make Unity not complain about it

stella3d: @mtschoen-unity made the comment clearer

AndrewTHEManeri:

PR #260, opened 2017-07-25
make test setup possible by eliminating static EVR contructor

stella3d:
move the code from EditorVR's static constructor to inside the Awake block to allow for setup within the context of tests.  

Cleaned up version of what we mobbed earlier - decided a separate function to keep all that block in was cleaner.

### DIFF
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -99,8 +99,8 @@ namespace UnityEditor.Experimental.EditorVR.Core
 			EditorPrefs.DeleteKey(k_SerializedPreferences);
 		}
 
-		// code from the previous static contructor moved here to allow for testability
-		void HandleInitialization()
+		// Code from the previous static constructor moved here to allow for testability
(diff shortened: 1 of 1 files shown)


Scripts/Core/EditorVR.cs
```diff
@@ -97,14 +99,44 @@ static void ResetPreferences()
 			EditorPrefs.DeleteKey(k_SerializedPreferences);
 		}
 
+		// code from the previous static contructor moved here to allow for testability
```
mtschoen: spelling on constructor. Also please capitalize the comment

Scripts/Core/EditorVR.cs
```diff
@@ -97,14 +99,44 @@ static void ResetPreferences()
...
+		void HandleInitialization()
```
mtschoen: Please make static

stella3d:
fixes made !

AndrewTHEManeri:

PR #261, opened 2017-07-27
core test setup fixture + tests for EditingContextManager

stella3d:
### Purpose of this PR

Adds unit testing for the functionality of `EditingContextManager`

Writing these tests, and tests for anything in EditorXR's core, requires some setup, which this also adds.  The `SetUpFixture` handles opening & closing of the EVR window without actual test code having to worry about it.

### Testing status

I've tested that the setup / teardown process happens without introducing any errors to the tests.  Sometimes the EVR window might fail to close when it's done, or a stray warning about trying to use an object that's been destroyed might pop up after the tests run, but these things don't affect the test results.  

Other than that, smoke test shows EVR still works as expected (tested with oculus, but shouldn't matter)

### Technical risk

_Low_ : code changes outside of tests amount to exposing some previously `private` members as `internal` to allow for more practical testability. 

### Comments to reviewers

One question i have on this is regarding exposing currently `private` member variables such as `m_DefaultToolStack` as `internal`- would it be preferable to leave them as they are & expose them like this, such as how i did `currentContext` ? It's a little more verbose, but i don't have a preference either way as long as we're consistent.
```csharp
internal List<MonoScript> defaultToolStack
{
	get { return m_DefaultToolStack; }
	set { m_DefaultToolStack = value; }
}
```


Tests/Editor/Unit/Core/EditingContextManagerTests.cs
```diff
@@ -0,0 +1,171 @@
...
+			var beginningContext = manager.currentContext;
+			manager.SetEditingContext(context2);
+			Assert.AreNotEqual(beginningContext, manager.currentContext);
+
+			manager.RestorePreviousContext();
+			Assert.AreEqual(beginningContext, manager.currentContext);
+		}
+
+		[Test]
+		public void LoadProjectSettings_IfAssetFound()
```
AndrewTHEManeri: Is this going to blow away our existing context manager settings if we run this on our local box?

stella3d: totally will - i'm going to write something into the test fixture setup to backup and restore

stella3d: test fixture now preserves the state of your settings at the start of the test

AndrewTHEManeri:
@mtschoen-unity What say you?

Tests/Test Helpers/Editor/EditorVRTestInitializer.cs
```diff
@@ -0,0 +1,35 @@
...
+			userSettingsBackup = EditingContextManager.LoadUserSettings();
+
+			EditingContextManager.ShowEditorVR();
+		}
+
+		[OneTimeTearDown]
+		public void CleanupAfterAllTests()
+		{
+			EditingContextManager.SaveProjectSettings(projectSettingsBackup);
+			EditingContextManager.SaveUserSettings(userSettingsBackup);
```
mtschoen-unity: Is this necessary yet? I'm wondering if it's possible to fake the settings so that we don't actually touch user settings. Not sure if that would be more trouble than it's worth.

Tests/Editor/Unit/Core/EditingContextManagerTests.cs
```diff
@@ -0,0 +1,171 @@
...
+		[Test]
+		public void LoadUserSettings_NewerThanProjectSettings()
+		{
+			EditingContextManager.SaveUserSettings(newSettings);
+			var loaded = EditingContextManager.LoadUserSettings();
+			Assert.AreEqual(newSettings.defaultContextName, loaded.defaultContextName);
+		}
+
+		[Test]
+		public void LoadUserSettings_OlderThanProjectSettings()
```
mtschoen-unity: What exactly are these two testing?

Tests/Test Helpers/Editor/EditorVRTestInitializer.cs
```diff
@@ -0,0 +1,35 @@
...
+		EditingContextManagerSettings projectSettingsBackup;
+		EditingContextManagerSettings userSettingsBackup;
+
+		[OneTimeSetUp]
+		public void SetupBeforeAllTests()
+		{
+			projectSettingsBackup = EditingContextManager.LoadProjectSettings();
+			userSettingsBackup = EditingContextManager.LoadUserSettings();
+
+			EditingContextManager.ShowEditorVR();
```
mtschoen-unity: I suppose this means any of these tests will cover whether EVR can start without errors?

Scripts/Proxies/TouchProxy.cs
```diff
@@ -21,7 +21,8 @@ public override IEnumerator Start()
 #if UNITY_EDITOR
 			EditorApplication.delayCall += () =>
 			{
-				transform.localPosition = Vector3.zero;
+				if (this != null)
+					transform.localPosition = Vector3.zero;
```
mtschoen-unity: Do we need to do the same for ViveProxy?

stella3d: maybe !

Tests/Test Helpers/Editor/EditorVRTestInitializer.cs
stella3d: sounds like a lot of trouble imho

Tests/Editor/Unit/Core/EditingContextManagerTests.cs
stella3d: we defer to project settings if they are newer than user settings

Scripts/Proxies/TouchProxy.cs
AndrewTHEManeri: ViveProxy isn't setting transform position anywhere in its start function

AndrewTHEManeri:
manual merge successful, performing github merge

PR #267, opened 2017-08-14
ObjectUtils test cleanup

stella3d:
the github diff view is wonky because i normalized tabs /  spaces while doing this, but the important part is that all created objects are now destroyed inside the tests, regardless of errors.


stella3d:
i thought about doing it that way.  I don't particularly see an advantage to it - you'd have to manually add stuff to the list every time.

mtschoen-unity:
Yeah, but you wouldn't have to keep adding variables for keeping track of more and more objects as you add more tests. I'm not sure if that's a real concern, though.

stella3d:
yeah - that's why i tried to move all the objects that are created during actual tests to just use `currentObject`, except the bounds tests as they use two at once.

stella3d:
working on these changes, but i don't think the double list makes sense.  It also makes it so you need to manage adding to a list and working with the list each time instead of just having the current object you're working on.

AndrewTHEManeri:
What about adding a tag to objects you create in the test, and then in cleanup just finding all the objects with that tag?

mtschoen-unity:
Actually, this isn't exactly what I was thinking. I'm also happy to make the changes on my end. Two other things I just noticed:

I wasn't paying attention to C# conventions in the tests, but we will need to change variable names to m_Something, etc.
Also, why use a custom delta/epsilon in GetBounds_Array? Vector3 == already compares with an epsilon.

mtschoen-unity:
Discussed changes w/ @stella3d and just made them myself. Didn't get rid of custom delta since we decided that Is.Equal.Within gives a better error message

AndrewTHEManeri:

PR #269, opened 2017-08-15
Menu hiding and other bugs

### LINKED ISSUE #65: Hide main menu when over any UI

### LINKED ISSUE #134: Don't allow two main menus at once

### LINKED ISSUE #190: Primitive tool shows gizmo on creation if selection cone leaves primitive volume
If while creating a primitive, the selection cone leaves the volume, the gizmo appears. 
(Incidentally, the gizmo resizes very aggressively if you move around the object.)
Gizmo should appear once trigger is released, not during creation/drawing stage, regardless of where the controller is.

### LINKED ISSUE #194: Increase chessboard height by 1/3
Just feels like it's not high enough now.

### LINKED ISSUE #197: Workspaces from previous session still in scene but don't follow me
https://cl.ly/kWvU

Here they are after I scaled the world, sitting about three meters above me. They're stuck there; they no longer respect world scaling or my movement.

### LINKED ISSUE #200: Raycast on UI overriding selection of objects
When near a workspace, the ray passes through an object I'm trying to select and casts on the UI.

Overall solution--the ray should stop when it hits something, and *not* pass through any object or UI.

### LINKED ISSUE #201: Radial appearing on opposite hand with direct selection

### LINKED ISSUE #208: Main menu not disappearing when an object is direct selected and moving second controller in (eg for scaling)

### LINKED ISSUE #220: Use selection outline color for highlight/rays

### LINKED ISSUE #253: World scale/multi-select changes
- Change world scaling back to grip trigger on Rift, grips on Vive
- Move multi-select to double-click on grip trigger (R) or thumbpad (V)
- Double click to enter, double-click to exit

### LINKED ISSUE #256: Duplicate Tooltip appears when hovering objects if BaseActions implement ITooltip
If any BaseAction (existing or new) implements ITooltip, a duplicate tooltip will appear near the origin when hovering over anything in EVR.

What appears to be happening is that the MultiRayInputModule checks the EVR object and passes this onto the TooltipModule.  It then checks if EVR has a component which implements ITooltip - since EVR contains all the actions as components, this check will return true pretty much all the time if any action implements ITooltip.  From there, it ends up spawning the duplicate tooltip.

mtschoen-unity:
### Purpose of this PR

Hide menus when hovering over UI... and then it snowballed. I'll break it down by issue first, then capture the changes not in issues.

Fixes #256: The issue here was that if anything on the EditorVR object itself (i.e. MainMenu) implemented ITooltip, its tooltip would always show up because every piece of UI is a child of the EditorVR object, and it gets hit when the hover events bubble up the hierarchy. The fix here (0a2e7cf2) was to just check if the object that comes into TooltipModule on hover is not the same object that it is attached to (i.e. EditorVR)

Fixes #253: Puts crawl back on grip and does double-click for multi-select. One funny side-affect of the new implementation is that when you toggle multi-select for one person in SceneFusion, everyone's state is toggled, because all of the SelecitonTools go in the linkedObjects list.  Because multi-select and crawl/worldscale on Rift overlap, there is extra code to delay the crawl interaction until a minimum distance or the double-click interval is exceeded (12ce37dd)

Fixes #220: This removes the one-color-per-hand feature and sources the Scene/Selected Outline setting for the ray/highlight color.

Fixes #208: Does what it says (0b7f2591)

Fixes #201: Does what it says (cf88cb84)

Fixes #200: This one has been bugging me for so long. The solution is to check the distance to the first scene object within isRayValid, and invalidate the UI ray if the scene ray distance was shorter. (b92ee4cc) This means that we need to actually let the UI hover happen (which was happening anyway) but issue hoverexit events if the ray is invalidated (685b9b03)

Fixes #197: From what I can tell, removing the Recenter in VRView fixes this problem, and still allows for our clever solution of rotating the view to face the same way you were facing the moment you quit EVR or took off the HMD. The recenter was causing the head-to-room offset to change every time you started EVR in a Rift, which causes the unpredictable offsets from workspaces. (224a4ae0)

Fixes #194: Does what it says (20ff737f)

Fixes #190: Does what it says (54c1f4c6)

Fixes #134: Does what it says (31ea47bb)

Fixes #65: Oh yeah... the reason we started this whole thing (d176d6a0). Obviously this evolved a bit, but the crux of it is that we have a method in EVR.Menus which is checked in RaycastSource.isValid. It is mostly there for the purpose of notifying the menu system of ray hovers, but can also be used to block the UI when the menu is being interacted with and shouldn't be hidden (939025b5)

Menus will hide when:
- Menu hand's ray hovers any UI, and distance is less than ~30cm
  - Hides after a 125ms delay, shows after a 250ms delay
- Menu bounds touch a workspace collider
  - Also disables activator button, even if menu is hidden
- Menu hand gets within 25cm of a hand which is direct manipulating (for two handed scale)
- Menu is activated on other hand

The menu will not hide if it is receiving raycasts (from the other hand)

The UI hover, Workspace, and Direct Selection cases are "temporary" states which will incur the auto-hide delay, and will automatically be cleared unless they are set every frame. This was because we don't get notified when the ray is hovering over no UI, so we would not be able to wait for a "not over UI anymore" call.

And now for the stuff that's not covered in issues:
Changes to how Ray Visibility is handled. We had a brief sync on this, but here's what I ended up with:
- We assume that the default state is visible, and if nothing has requested visibility changes, the ray and cone are both visible
- Classes that implement IRayVisibilitySettings can add/remove settings "requests." When they call AddRayVisibilitySettings they provide their own instance, a bool for ray/cone visibility, and a priority (int). This will override any existing settings completely, regardless of priority.
- Every frame, DefaultProxy ray selects the visibility settings with the highest priority and applies them. If more than one object requested settings at the same priority, the bools are combined using &. In other words, if any objects requested a "hide" with the top priority, they'll get it. If you need to override existing settings to show the ray you _must_ use a higher priority.
- Implementors can also remove their settings by calling RemoveRayVisibilitySettings with their instance. All settings are removed.

The issue this is fixing is crawling while hovering a direct selection. What happens is that the TransformTool is trying to hide the ray, but not the cone. Then the LocomotionTool comes in and tries to hide the cone, but can't, since TransformTool has locked the ray. Thus, we need a way to let LocomotionTool override TransformTool. In this case, the "visible if no settings" rule is enough, and we do not need to mess with priority. Currently, no system needs a priority > 0. This system solidified in d3271cc3. I played around with setting up this interface to take a generic Settings object rather than cone/ray/priority in case we ended up having rays with different kinds of settings, but abandoned the idea since I couldn't come up with a good way to let users plug in new settings class types (7552e14).

Related to this issue, I split the rendering from the physics in the DefaultProxyRay prefab in order to leave the cone IntersectionTesters active while we hide the cone visuals. This is so that we can continue to get direct selection information while the cones are hidden. The bug this fixes is a situation where, when you release you hand from crawling while direct selecting (as above), the ray would start to show for a split second while the cones were scaling in, because we didn't "know" we had a direct selection until the cone got back (f7aea098). We now manually disable direct hovers and direct manipulation based on cone visibility (6e7d7748).

I made a bunch of fixes to the radial menu because they were bugging me (79b7327c, 48b52619)

The ray now hides when you grab items from the project view (a83c0d48)

Various things (workspace front panel, workspace top panel, menu pane, expand radial buttons) had to become raycasts targets to trigger hiding the menu (03782a8d, etc)

I eliminated DirectSelectInput/DirectSelectData because we didn't need it anymore (TransformTool now checks the trigger) (b46f4a6b)

### Testing status

I've been testing this throughout development, and we have done multiple design reviews. The SIGGRAPH presentation also happened on top of this branch. I've also run successful automated tests on the current state.

### Technical risk

_Medium_ : This touches lots of things and refactors major systems.

### Comments to reviewers

Remember to clear your EditorVR settings (Preferences -> EditorVR -> Reset to Defaults) to clear out old workspace bounds which are too tall (in Y) and will hide the menus prematurely.

As for why this is such a huge PR: The "hiding the menus" bug ended up being a proxy for general usability, and it was necessary to apply all of these fixes in the same environment to "see how it feels" in design reviews and while considering possible solutions.


mtschoen:
208 - Are you talking about two-handed scaling in general? Like you should be able to two-handed scale without touching the object? I have been wanting to tweak the two-handed scaling to either include rotation or change how it decides to "push away" the object from you as you scale so it's easier to keep grabbing with the second hand. Ideally, if you've poked the cone from your second hand into the object, it should stay poked into that surface while you scale. I figured this was outside the scope of this PR.

201 - Confused by what you mean here. The idea is that when you grab an object with the hand with the radial menu, it kicks over to the other hand.  It would only be open in the first place if you ray-select something. Can you explain what you're seeing?

65 - What did you mean by "stick your arm into the menu?" The other controller shouldn't have anything to do with hiding the menu. It should keep the menu open as long as the other ray is hovering it.

Interesting point about selected objects. Certainly all scene objects shouldn't count. And the manipulator is UI. Why would you want the selected object to hide the menu?

The opacity state for temporary hiding is something Dylan and I discussed, as well, and decided it wasn't important enough to block this PR. Thankfully, we clearly distinguish a temporary-hidden state from permanent menu hiding, so we can easily use this to distinguish visibility states.

mtschoen-unity:
208 - Are you talking about two-handed scaling in general? Like you should be able to two-handed scale without touching the object? I have been wanting to tweak the two-handed scaling to either include rotation or change how it decides to "push away" the object from you as you scale so it's easier to keep grabbing with the second hand. Ideally, if you've poked the cone from your second hand into the object, it should stay poked into that surface while you scale. I figured this was outside the scope of this PR.

201 - Confused by what you mean here. The idea is that when you grab an object with the hand with the radial menu, it kicks over to the other hand. It would only be open in the first place if you ray-select something. Can you explain what you're seeing?

65 - What did you mean by "stick your arm into the menu?" The other controller shouldn't have anything to do with hiding the menu. It should keep the menu open as long as the other ray is hovering it.

Interesting point about selected objects. Certainly all scene objects shouldn't count. And the manipulator is UI. Why would you want the selected object to hide the menu?

The opacity state for temporary hiding is something Dylan and I discussed, as well, and decided it wasn't important enough to block this PR. Thankfully, we clearly distinguish a temporary-hidden state from permanent menu hiding, so we can easily use this to distinguish visibility states.

AndrewTHEManeri:
208 - Yeah.  For example, I was scaling one of the low poly trees.  At a certain size (not too gigantic) that it becomes unpredictable if it is going to detect the second hand as scaling.  Some kind of extra check such as the second controller ray hitting the object if the first is already pulling it around would probably be enough.  And yes - this does not have to be in the scope of this PR.  I'll make an issue.
201 - Not able to repro today.  If I see it again I'll record a video.
65 - Two images here.  First is the hand example
image
That will eventually turn into a direct selection cone if you aim it exactly right, but you are essentially aiming blind.  
Second is the selected object example.  I agree not all scene objects should count.  But in this scenario:
image
I think it would be much more useful if that turned back into the direct selection cone.

mtschoen-unity:
Ah I see what you mean. So what's going on in that first image there is that the Activator is a piece of UI, and the invisible ray that's still coming out of the rayorigin has to hit it to hide the menu. We could probably check specifically if the menu overlaps the activator collider. Thoughs, @timoni?

In the second one, is this what you mean about the selected object hiding the menu? Is this just for the purpose of direct selection? I suppose that we could add a rule for "if you direct-hover the selected object, hide the menu."

AndrewTHEManeri:
Re: The second one - That sounds like it would work.  I think in general the scenario where significant amount of object geometry is intersecting a menu is one we should work to avoid.  Users trying to select something in the menu are just as likely to select whatever it is that is intersecting with it.

mtschoen-unity:
Yeah. That's just a tough thing to test. At least we can keep testing the cones now that they're hidden, so that helps. I can give this a shot and see how disruptive the code changes will be. I'll DM you on slack before I push anything

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -441,22 +458,51 @@ internal void PreProcessRaycastSource(Transform rayOrigin)
...
+					if (dpr)
 					{
-						node = deviceData.node;
-						break;
+						return dpr.GetColor();
 					}
 				}
 
-				return node;
+				return default(Color);
```
AndrewTHEManeri: Not appropriate to look this up from the highlightModule like in set?

Scripts/Interfaces/Capability/ITwoHandedScaler.cs
```diff
@@ -0,0 +1,19 @@
...
+	/// <summary>
+	/// Provides a method used to check the status of two-handed scaling
+	/// </summary>
+	public interface ITwoHandedScaler
+	{
+		/// <summary>
+		/// Returns whether the given ray origin is involved in two-handed scaling
+		/// </summary>
+		/// <param name="rayOrigin">The ray origin to check</param>
+		/// <returns></returns>
```
AndrewTHEManeri: I know its redundant but returns field should be fille din

Scripts/Modules/HierarchyModule.cs
```diff
@@ -28,10 +28,12 @@ void Awake()
 				m_IgnoreList.Add(manager.gameObject);
 			}
 
+#if UNITY_EDITORVR
```
AndrewTHEManeri: compilation test need?

Scripts/Modules/HighlightModule/HighlightModule.cs
```diff
@@ -28,19 +28,24 @@ sealed class HighlightModule : MonoBehaviour, IUsesGameObjectLocking
...
 		void Awake()
 		{
 			s_BakedMesh = new Mesh();
+
+			if (EditorPrefs.HasKey(k_SelectionOutlinePrefsKey))
+			{
+				var selectionColor = MaterialUtils.PrefToColor(EditorPrefs.GetString(k_SelectionOutlinePrefsKey));
+				selectionColor.a = 1;
+
+				m_RayHighlightMaterial = Instantiate(m_RayHighlightMaterial);
```
AndrewTHEManeri: I think we should always instantiate - otherwise this material is inconsistently what the field has referenced or an instance of it

Scripts/Modules/IntersectionModule/IntersectionModule.cs
```diff
@@ -74,9 +74,9 @@ void Update()
 				{
 					var intersectionFound = false;
 					m_Intersections.Clear();
-					if (m_SpatialHash.GetIntersections(m_Intersections, tester.renderer.bounds))
+					if (m_SpatialHash.GetIntersections(m_Intersections, tester.collider.bounds))
```
AndrewTHEManeri: What does switching from renderer to collider gain/cost us?

Scripts/Interfaces/Entity/IMainMenu.cs
```diff
@@ -34,6 +34,21 @@ public interface IMainMenu : IMenu, IUsesMenuActions, ISelectTool
...
 		/// </summary>
 		Transform targetRayOrigin { set; }
+
+		/// <summary>
+		/// The local bounds of this menu
+		/// </summary>
+		Bounds localBounds { get; }
+
+		/// <summary>
+		/// Is the UI pointer hovering over the menu?
```
mtschoen-unity: Should I call the property "focus"?

AndrewTHEManeri: Yeah, that sounds good

Scripts/Interfaces/Entity/IMenu.cs
```diff
@@ -22,6 +22,11 @@ public interface IMenu
 		/// Root GameObject for visible menu content
 		/// </summary>
 		GameObject menuContent { get; }
+
+		/// <summary>
+		/// If the rayOrigin this menu is attached to is hovering UI, hide it if the raycast distance is less than this
+		/// </summary>
+		float hideDistance { get; }
```
mtschoen-unity: Hm. Since I only needed a distance along the ray, I kept it to a single float. I was also not a huge fan of something so implementation-specific. Exposing localbounds is a good idea, though it does still contain the expectation that they're local to "something."

Something else to consider is whether this belongs in public interfaces or if we want to handle our "special case" main menu that's attached to your hand separately from other menu features.

The concept of auto-hiding the menu when it intersects with things or an associated ray is hovering UI is conceptually coupled to the idea of "menu on rayorigin," so I figured it was safe to assume at this level.

TL;DR: I think just exposing localbounds makes sense in the scope of this PR.

Menus/MainMenu/MainMenu.cs
```diff
@@ -273,14 +277,9 @@ void OnButtonHovered(Transform rayOrigin)
 			this.Pulse(this.RequestNodeFromRayOrigin(rayOrigin), m_ButtonHoverPulse);
 		}
 
-		void OnOpening()
-		{
-			this.Pulse(node, m_ShowPulse);
-		}
-
-		void OnClosing()
+		public void SendVisibilityPulse()
```
mtschoen-unity: Yeah. The issue is that we need to distinguish between a temp-hidden and perma-hidden state because we specifically needed to cut haptics on auto-hide.

I could just add a "withHaptics" argument or perhaps "isTemporary" if we want to go with that semantic.

Menus/RadialMenu/RadialMenu.cs
```diff
@@ -83,6 +83,8 @@ public bool visible
 
 		public Node? node { get; set; }
 
+		public float hideDistance {get { return Mathf.Infinity; } }
```
mtschoen-unity: The idea is that the menu should always hide on any UI hover, rather than the distance-based check on mainmenu. Since the radial menu never extends beyond the rayorigin, the distance-based check doesn't apply.

To be honest, I don't notice the radial menu showing/hiding since it's so transparent. I wasn't sure if "never hide" or "hide on every hover regardless of distance" was the right default.

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -197,119 +109,282 @@ internal void UpdateMenuVisibilities()
...
 				{
-					var deviceData = m_ActiveDeviceData[i];
 					var alternateMenu = deviceData.alternateMenu;
 					var mainMenu = deviceData.mainMenu;
 					var customMenu = deviceData.customMenu;
 					var menuHideFlags = deviceData.menuHideFlags;
 
-					// Move alternate menu to another device if it conflicts with main or custom menu
-					if (alternateMenu != null && (menuHideFlags[mainMenu] == 0 || (customMenu != null && menuHideFlags[customMenu] == 0)) && menuHideFlags[alternateMenu] == 0)
+					var mainMenuVisible = mainMenu != null && (menuHideFlags[mainMenu] & MenuHideFlags.Hidden) == 0;
```
mtschoen-unity: Oof. Yeah, it's hard to know where to draw the line. Anything I could think to comment was a "what, not a why," so I opted to not write it.  I'll take another look and try to add comments or improve clarity.

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -197,119 +109,282 @@ internal void UpdateMenuVisibilities()
...
+								m_WorkspaceComponents.Clear();
+								overlap.GetComponents(m_WorkspaceComponents);
+								if (m_WorkspaceComponents.Count > 0)
+									hoveringWorkspace = true;
+							}
 						}
+					}
 
-						menuHideFlags[alternateMenu] |= MenuHideFlags.Hidden;
+					var menus = menuHideFlags.Keys.ToList();
```
mtschoen-unity: Ah. I thought I had gotten to this. I think the solution is just to do a normal foreach (kvp in dictionary) instead of Keys.ToList().

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -197,119 +109,282 @@ internal void UpdateMenuVisibilities()
...
+						var alternateMenuVisible = alternateMenu != null && (menuHideFlags[alternateMenu] & MenuHideFlags.Hidden) == 0;
 
-						var customMenu = deviceData.customMenu;
-						if (customMenu != null)
-							menuHideFlags[customMenu] &= ~MenuHideFlags.Hidden;
+						deviceData.menuAutoShowTimes[mainMenu] = 0;
+
+						if (deviceData.rayOrigin == rayOrigin)
+						{
+							menuHideFlags[mainMenu] ^= MenuHideFlags.Hidden;
```
mtschoen-unity: yep. Should I comment? This is one of the few things that hasn't changed from the previous implementation :)

Menus/MainMenu/MainMenu.cs
AndrewTHEManeri: I like the temporary arg

Scripts/Modules/IntersectionModule/IntersectionModule.cs
mtschoen-unity: They're no longer on the same object. The collider is actually new, so it will technically cost us its bounds computation and whatever physics overhead we get from moving static colliders around in edit mode (really not sure what's involved on that one).

It's necessary because we need direct hover info even while the cones are invisible, and wouldn't be giving us proper bounds information (scaled down and probably not updated when inactive).

There is now a MeshCollider with the cone as its mesh that is used for giving us the AABB for coarse testing as well as the mesh reference for testing each ray.

Scripts/Modules/HierarchyModule.cs
mtschoen-unity: Actually I added this because the compilation test failed. Was probably introduced by my hierarchy update from a few weeks ago.

Menus/RadialMenu/RadialMenu.cs
AndrewTHEManeri: I would probably leans towards never hide since would technically be doing less but we should probably set up some kind of test scene where a user is surrounded by UI and see how both feel.

Scripts/Interfaces/Capability/ITwoHandedScaler.cs
mtschoen-unity: Good catch. Missed that one.

Scripts/Core/EditorVR.Rays.cs
mtschoen-unity: Hm. I think I do this because this gets called once or twice before highlightModule is instantiated. I'll check it out.

Scripts/Core/EditorVR.Menus.cs
AndrewTHEManeri: Might as well, don't see many xors in this code

Scripts/Core/EditorVR.Rays.cs
AndrewTHEManeri: Ah, yeah, I've been thinking about that situation with func. injection in general

Scripts/Modules/HierarchyModule.cs
AndrewTHEManeri: Sounds good

Scripts/Modules/IntersectionModule/IntersectionModule.cs
AndrewTHEManeri: Got it, thanks

Workspaces/Base/Workspace.cs
```diff
@@ -9,12 +9,13 @@
...
 {
-	abstract class Workspace : MonoBehaviour, IWorkspace, IInstantiateUI, IUsesStencilRef, IConnectInterfaces, IUsesViewerScale, IControlHaptics, IRayToNode
+	abstract class Workspace : MonoBehaviour, IWorkspace, IInstantiateUI, IUsesStencilRef, IConnectInterfaces,
+		IUsesViewerScale, IControlHaptics, IRayToNode
 	{
 		const float k_MaxFrameSize = 100f; // Because BlendShapes cap at 100, our workspace maxes out at 100m wide
 
 		public static readonly Vector3 DefaultBounds = new Vector3(0.7f, 0.4f, 0.4f);
-		public static readonly Vector3 MinBounds = new Vector3(0.55f, 0.4f, 0.1f);
+		public static readonly Vector3 MinBounds = new Vector3(0.55f, 0f, 0.1f);
```
AndrewTHEManeri: is 0 on the y a safe min value here?

Workspaces/Base/Workspace.cs
```diff
@@ -77,11 +83,11 @@ public Bounds outerBounds
 		{
 			get
 			{
-				const float kOuterBoundsCenterOffset = 0.225f; //Amount to lower the center of the outerBounds for better interaction with menus
-				return new Bounds(contentBounds.center + Vector3.down * kOuterBoundsCenterOffset,
+				const float kOuterBoundsCenterOffset = 0.09275f; //Amount to extend the bounds to include frame
```
AndrewTHEManeri: k_

Workspaces/Base/Workspace.cs
```diff
@@ -142,12 +148,27 @@ public virtual void Setup()
...
 
+			m_OuterCollider = gameObject.AddComponent<BoxCollider>();
+			m_OuterCollider.isTrigger = true;
+			m_FrameCollider = gameObject.AddComponent<BoxCollider>();
+			m_FrameCollider.isTrigger = true;
+
+			var startingBounds = m_CustomStartingBounds ?? DefaultBounds;
 			//Do not set bounds directly, in case OnBoundsChanged requires Setup override to complete
-			m_ContentBounds = new Bounds(Vector3.up * DefaultBounds.y * 0.5f, m_CustomStartingBounds ?? DefaultBounds); // If custom bounds have been set, use them as the initial bounds
+			m_ContentBounds = new Bounds(Vector3.up * startingBounds.y * 0.5f, startingBounds); // If custom bounds have been set, use them as the initial bounds
```
AndrewTHEManeri: Does the center need to be set here?

Workspaces/Base/Workspace.cs
```diff
@@ -267,6 +292,10 @@ IEnumerator AnimateResetSize()
 		public virtual void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
 		{
 			m_WorkspaceUI.ProcessInput((WorkspaceInput)input, consumeControl);
+
```
AndrewTHEManeri: Is this changing every time input is processed?

Workspaces/MiniWorldWorkspace/MiniWorldWorkspace.cs
```diff
@@ -106,9 +105,10 @@ public float zoomSliderMax
 
 		public override void Setup()
 		{
+			const float yBounds = 0.5f;
 			// Initial bounds must be set before the base.Setup() is called
-			minBounds = new Vector3(MinBounds.x, MinBounds.y, 0.25f);
-			m_CustomStartingBounds = new Vector3(MinBounds.x, MinBounds.y, 0.5f);
+			minBounds = new Vector3(MinBounds.x, yBounds, 0.25f);
```
AndrewTHEManeri: Why set to a const here?

Scripts/UI/MainMenuActivator.cs
```diff
@@ -68,47 +80,72 @@ public bool activatorButtonMoveAway
 		Coroutine m_ActivatorMoveCoroutine;
 		Vector3 m_AlternateActivatorLocalPosition;
 
+		bool m_Disabled;
+		Material m_IconMaterial;
+		Color m_EnabledColor;
```
mtschoen-unity: We get enabledColor from the color the in the material. I could switch this to specifying both in the script, but generally we've been trying to avoid having a value serialized in two places, even in situations like this.

Scripts/Proxies/DefaultProxyRay.cs
```diff
@@ -139,24 +91,66 @@ public void SetLength(float length)
 
 		public void SetColor(Color c)
 		{
+			m_LineRenderer.SetColors(c, c);
```
mtschoen-unity: Hm. I believe this is how it used to work, as well. That could also explain why the line renderer was giving us that non-clamped alpha issue. I will see if I can safely remove this.

Scripts/Modules/TooltipModule/TooltipModule.cs
```diff
@@ -177,14 +188,27 @@ void UpdateVisuals(ITooltip tooltip, TooltipUI tooltipUI, Transform target, floa
...
+				ShowTooltip(tooltip);
+		}
+
+		public void OnRayHovering(GameObject gameObject, RayEventData eventData)
+		{
+			if (gameObject == this.gameObject)
+				return;
+
 			var tooltip = gameObject.GetComponent<ITooltip>();
 			if (tooltip != null)
```
mtschoen-unity: This came up for the "Main Menu is Hidden" tooltip on the MainMenuActivator. Because its state can change from no-tooltip to valid tooltip while you are hovering it, we need to check that. In this case, we have a null tooltip string up until the menu is hidden.  Same thing applies to checking while it's being shown whether the tooltip is still valid. If the menu moves away from a workspace while you're hovering the activator, the tooltip should hide, even though the ray is still hovering it.

Workspaces/MiniWorldWorkspace/MiniWorldWorkspace.cs
mtschoen-unity: This was to increase the height of the chessboard. It's using the default minbounds, but we want the height to be a little bigger, hence the new const.

Workspaces/Base/Workspace.cs
mtschoen-unity: Yes. The transform of the workspace has its y = 0 point at the top of the frame. The content extends out from there, so as it grows in height, the center goes up by half the height.

Workspaces/Base/Workspace.cs
mtschoen-unity: Good point. I can make this a part of WorkspaceUI.

Workspaces/Base/Workspace.cs
mtschoen-unity: Hm. Should I also change the yBounds const in MiniWorldWorkspace? I realize that we've been inconsistent about const in the local method scope. Some have k, some don't. None of them thus far have had a k_

Workspaces/Base/Workspace.cs
mtschoen-unity: Yes. The frame itself has its own collider, so we do have something for workspaces with 0 height.

Tools/SelectionTool/SelectionInput.cs
```diff
@@ -9,5 +9,6 @@ public class SelectionInput : ActionMapInput {
 		
 		public ButtonInputControl @multiSelect { get { return (ButtonInputControl)this[0]; } }
 		public ButtonInputControl @select { get { return (ButtonInputControl)this[1]; } }
+		public ButtonInputControl @multiSelectAlt { get { return (ButtonInputControl)this[2]; } }
```
mtschoen-unity: No. At least not as far as I know.

We have one VRDevice which covers Vive and Oculus. In fact, it includes a Button2 which Vive doesn't have. Extending this to include more VRDevices was a TODO for when input stuff coalesced, but I think we also still want to default to a generic VR device so that you don't have to specify support for individual hardware.

Scripts/Utilities/MaterialUtils.cs
```diff
@@ -51,7 +52,7 @@ public static UnityMaterial[] CloneMaterials(Renderer renderer)
 		}
 
 		// from http://wiki.unity3d.com/index.php?title=HexConverter
-		// Note that Color32 and Color implictly convert to each other. You may pass a Color object to this method without first casting it.
+		// Note that Color32 and Color implicitly convert to each other. You may pass a Color object to this method without first casting it.
 		public static string ColorToHex(Color32 color)
 		{
 			var hex = color.r.ToString("X2") + color.g.ToString("X2") + color.b.ToString("X2");
```
mtschoen-unity: /shrug

Scripts/UI/MainMenuActivator.cs
```diff
@@ -68,47 +80,72 @@ public bool activatorButtonMoveAway
...
 
+			SetHighlight(false);
+		}
+
+		void SetHighlight(bool highlighted)
+		{
 			if (m_HighlightCoroutine != null)
 				StopCoroutine(m_HighlightCoroutine);
 
 			m_HighlightCoroutine = null;
```
mtschoen-unity: Not sure. @dunity wrote it. I can get rid of it, but I was just moving this code around.

Menus/MainMenu/MainMenu.cs
mtschoen-unity: Ah, right. I forgot a crucial thing here. When we're showing the menu, the hideflags are set to 0, so we don't know whether it was because of the activator or because of an auto-hide. That's why SendVisibilityPulse is called by the activator method. The only way I can think to solve this is to keep track of a lastHideFlags state for each menu, which seems awkward.

Thoughts?

Scripts/Core/EditorVR.Menus.cs
mtschoen-unity: Right. So this is a somewhat rare case where we're modifying each value-type value of a dictionary while iterating over it. We need to copy *something* in order to do this, because this doesn't work:

foreach (var kvp in dictionary)
    dictionary[kvp.Key] = someValue;

You get an InvalidOperationException.

My current solution is to keep a static list around and do this:

k_Menus.Clear();
k_Menus.AddRange(menuHideFlags.Keys);
foreach (var menu in k_Menus)
{
	menuHideFlags[menu] &= ~MenuHideFlags.Temporary;
}

This still allocates on KeyCollection.CopyTo, but it allocates less. I believe the only way to eliminate allocation entirely is to iterate twice, like so:

k_Menus.Clear();
foreach (var kvp in menuHideFlags)
{
	k_Menus.Add(kvp.Key);
}

foreach (var menu in k_Menus)
{
	menuHideFlags[menu] &= ~MenuHideFlags.Temporary;
}

This makes me think we should just keep a separate list of menus per device data, or maybe split the dictionary into two lists, and do a lookup only when needed? Curious if you have any alternate solutions in mind.

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -197,119 +109,282 @@ internal void UpdateMenuVisibilities()
...
 				}
+
+				evr.GetModule<DeviceInputModule>().UpdatePlayerHandleMaps();
 			}
 
-			internal static void OnMainMenuActivatorHoverEnded(Transform rayOrigin)
+			internal static bool IsValidHover(MultipleRayInputModule.RaycastSource source)
 			{
+				var go = source.draggedObject;
+				if (!go)
```
mtschoen-unity: This fix has been pushed, but I did !go on the line below, and it didn't mark the change.

Scripts/Core/EditorVR.Rays.cs
mtschoen-unity: Seems ok. Pushing a fix

Scripts/Proxies/DefaultProxyRay.cs
mtschoen-unity: hm... removing this fails to set the color properly. I don't have time to look into why right now, but I'll figure it out.

Scripts/Core/EditorVR.Menus.cs
mtschoen-unity: The fix here was to combine hideflags and hide/show time into a MenuHideData class, so now we can set fields on those classes and we're not modifying the dictionary.

It looks like that last snippet in my previous comment is the best/only way to modify dictionary values in a loop without allocating.

Scripts/UI/MainMenuActivator.cs
```diff
@@ -68,47 +80,72 @@ public bool activatorButtonMoveAway
...
 		public Transform rayOrigin { private get; set; }
 		public Transform menuOrigin { private get; set; }
 		public Node? node { get; set; }
 
-		public event Action<Transform> hoverStarted;
-		public event Action<Transform> hoverEnded;
 		public event Action<Transform, Transform> selected;
 
-		public void OnPointerEnter(PointerEventData eventData)
+		public bool disabled
```
mtschoen-unity: will change to interactable

Scripts/Modules/TooltipModule/TooltipModule.cs
mtschoen-unity: I just found another bug related to this tooltip which I couldn't find an unobtrusive solution for, and therefore probably will be introduced by this PR:

If you have a tool open, hovering the pinned tool button, which is a child of the main menu activator, you see both tooltips. I'd like a code solution, i.e. not letting tooltip activations "bubble up" past the first valid new tooltip, but this would involve major changes to the input module.

Also, this is all going to change in the toolflow PR.

mtschoen-unity:
@AndrewTHEManeri that wraps up the first (and hopefully only?) round of changes. Along the way, I fixed a null ref bug in ff1a504, and cleaning up UpdateMenuVisibilities to make more sense. Basically I realized as I was commenting stuff that it was unnecessary or so hard to explain there was probably a better way. Thanks for asking me to do that 👍

Menus/MainMenu/MainMenu.cs
```diff
@@ -97,11 +83,9 @@ public bool visible
 
 		public Transform rayOrigin { private get; set; }
 
-		public float hideDistance { get { return m_MainMenuUI.menuHeight; } }
-
 		public Bounds localBounds { get { return m_MainMenuUI.localBounds; } }
 
-		public bool hovering { get { return m_MainMenuUI.hovering; } }
+		public bool focus { get { return m_MainMenuUI.hovering; } }
```
AndrewTHEManeri: should MainMenuUI's variable also be changed to focus?

Scripts/Utilities/MaterialUtils.cs
```diff
@@ -105,6 +105,14 @@ public static Color RandomColor()
 			return new Color(r, g, b);
 		}
 
+		public static Color HueShift(Color color, float shift)
```
AndrewTHEManeri: Extension method?

Menus/MainMenu/MainMenu.cs
mtschoen-unity: I thought of that, but I figured that "hovering" is what's happening here. It makes sense that a menu might abstractly be "focused" by a number of things, but in the case of the UI, hovering is what's happening.

AndrewTHEManeri: Good, ok

Scripts/Utilities/MaterialUtils.cs
mtschoen-unity: I tried that at first but it's more code. Here's what an extension method would look like:
>public static void HueShift(this Color color, float shift)
>{
>    Vector3 hsv;
>    Color.RGBToHSV(color, out hsv.x, out hsv.y, out hsv.z);
>    hsv.x = Mathf.Repeat(hsv.x + shift, 1f);
>    var c = Color.HSVToRGB(hsv.x, hsv.y, hsv.z);
>    color.r = c.r;
>    color.g = c.g;
>    color.b = c.b;
>    color.a = c.a;
>}

AndrewTHEManeri: That still won't work because color is a struct - I was just assuming it would also return the color that was shifted, but I see that Vector3's don't do things that way so yes, your current method (non extension is best)

mtschoen-unity: Actually wouldn't that work? Since I'm setting .r, .g, etc. on the same color?

But yeah, simpler as a non-extension method.

AndrewTHEManeri: This parameters are still regular parameters, so the color being passed in is still getting copy-constructed (because of its struct-ness).
C# does not allow ref this parameters...yet

Tools/TransformTool/TransformTool.cs
```diff
@@ -703,6 +734,11 @@ void UpdateManipulatorAction()
 			m_ManipulatorToggleAction.tooltipText = isStandard ? "Switch to Scale Manipulator" : "Switch to Standard Manipulator";
 			m_ManipulatorToggleAction.icon = isStandard ? m_ScaleManipulatorIcon : m_StandardManipulatorIcon;
 		}
+
+		public bool IsTwoHandedScaling(Transform rayOrigin)
+		{
+			return m_Scaling && m_GrabData.Any(kvp => kvp.Value.rayOrigin == rayOrigin);
```
mtschoen-unity: If the first condition is false, the second doesn't get executed.

Scripts/Modules/TooltipModule/TooltipModule.cs
```diff
@@ -177,14 +188,27 @@ void UpdateVisuals(ITooltip tooltip, TooltipUI tooltipUI, Transform target, floa
 
 		public void OnRayEntered(GameObject gameObject, RayEventData eventData)
 		{
+			if (gameObject == this.gameObject)
```
mtschoen-unity: This was a fix for #256. I opened another issue (#278) for the general case, since it came back up again.

Scripts/Core/EditorVR.Rays.cs
```diff
@@ -441,22 +461,52 @@ internal void PreProcessRaycastSource(Transform rayOrigin)
...
 				{
-					var deviceData = evrDeviceData[i];
 					if (!deviceData.proxy.active)
 						continue;
 
 					if (deviceData.rayOrigin == rayOrigin)
+						return deviceData.node;
+				}
+
+				foreach (var kvp in evr.GetNestedModule<MiniWorlds>().rays)
```
mtschoen-unity: I believe that in almost every case you get an IEnumerator which does not call GetNestedModule again. @AndrewTHEManeri, do you have a more definitive answer? I could go write a test, but I'm mostly sure on this one.

Menus/MainMenu/MainMenu.cs
```diff
@@ -121,6 +109,25 @@ void Start()
...
+				if (m_MainMenuUI)
+				{
+					m_MainMenuUI.visible = visible;
+					if (!temporary)
+						SendVisibilityPulse();
+				}
+			}
+		}
+
+		public bool GetVisible()
```
mtschoen-unity: Hm. I don't really like having two ways of setting visibility. I'm fine with renaming GetVisible to a get-only property, but I don't think it is a good idea to have a set_visible as well as SetVisibilty with an extra parameter. I agree that it does break convention, but I think that's less important than DRY in this case. Also, in case you didn't catch the change, this used to be as it was, and I had added a SendVisibilityPulse method to IMainMenu, which was called directly in OnMainMenuActivatorSelected.

@AndrewTHEManeri, thoughts?

Scripts/Core/EditorVR.Rays.cs
AndrewTHEManeri: It won't, it will evaluate the function once, giving you the array, and then call the rays method on that.

AndrewTHEManeri: Yes, it will evaluate getnestedmodule once, giving you the miniworld, then evaluate rays once giving you the dictionary, then enumerate through that

Menus/MainMenu/MainMenu.cs
AndrewTHEManeri: This came about because I don't like such a specific implementation of haptic feedback making its way into the menu interface. 
Capturing the 'temporaryness' is important for later, however, especially if we start alpha-ing out menus instead of hiding them completely.
Another option is this doesn't take a bool, but some kind of flags enum that can encapsulate whether this is a major or minor(temporary) operation.

Scripts/Core/EditorVR.Rays.cs
mtschoen-unity: Cool. @dunity maybe you were thinking of the common pitfall with regular for loops where you can incur extra cost accessing the length of the collection?

Menus/MainMenu/MainMenu.cs
mtschoen-unity: Either way, the issue is that we're using methods instead of properties, which I don't think can be helped. @dunity, would you still like me to turn GetVisible into a set-only property? Should we leave it as it is?

AndrewTHEManeri: with the flags-enum route you could go back to a property and in fact a singular one

mtschoen-unity: Oh, I see what you mean. Hm... that still breaks the "single bool property" convention, unless we go to an enum for all UI visibilities. Would it make sense to just formalize MenuHideFlags and set that on the UI objects?

Maybe we should just have a chat about this tomorrow.

AndrewTHEManeri: chat sounds good

dunity: I'm for the flags property suggestion.  Best of both worlds imo.

Scripts/Core/EditorVR.Rays.cs
dunity: @mtschoen-unity Indeed I was.  Looks all good then.

Scripts/Modules/TooltipModule/TooltipModule.cs
dunity: Ah, thanks for the clarity.

mtschoen-unity:
Just got an exception while testing. I tried to close the main menu on the opposite hand after messing around with the tool menu. Not exactly sure what caused it. Here's a stack trace:

> ArgumentOutOfRangeException: Argument is out of range.
Parameter name: index
System.Collections.Generic.List`1[UnityEditor.Experimental.EditorVR.IPinnedToolButton].get_Item (Int32 index) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/List.cs:633)
UnityEditor.Experimental.EditorVR.Menus.PinnedToolsMenuUI.DeleteHighlightedButton () (at Assets/EditorVR/Menus/PinnedToolMenu/Scripts/PinnedToolsMenuUI.cs:417)
UnityEditor.Experimental.EditorVR.Menus.PinnedToolButton.OnSecondaryButtonClicked () (at Assets/EditorVR/Menus/PinnedToolMenu/PinnedToolButton/PinnedToolButton.cs:505)
UnityEditor.Experimental.EditorVR.UI.GradientButton.OnPointerClick (UnityEngine.EventSystems.PointerEventData eventData) (at Assets/EditorVR/Scripts/UI/GradientButton.cs:536)
UnityEngine.EventSystems.ExecuteEvents.Execute (IPointerClickHandler handler, UnityEngine.EventSystems.BaseEventData eventData) (at C:/buildslave/unity/build/Extensions/guisystem/UnityEngine.UI/EventSystem/ExecuteEvents.cs:50)
UnityEngine.EventSystems.ExecuteEvents.Execute[IPointerClickHandler] (UnityEngine.GameObject target, UnityEngine.EventSystems.BaseEventData eventData, UnityEngine.EventSystems.EventFunction`1 functor) (at C:/buildslave/unity/build/Extensions/guisystem/UnityEngine.UI/EventSystem/ExecuteEvents.cs:261)
UnityEditor.Experimental.EditorVR.Core.EditorVR:Update()

mtschoen-unity:
I've also been seeing this warning pretty consistently:

> New positions does not match size of existing array.  Adjusting vertex count as well
UnityEngine.Debug:LogWarning(Object)
VRLineRenderer:SetPositions(Vector3[], Boolean) (at Assets/EditorVR/libs/VRLineRenderer/Scripts/VRLineRenderer.cs:185)
UnityEditor.Experimental.EditorVR.Core.EditorVR:Update()

mtschoen-unity:
As for general design feedback: The haptic pulse when you initiate spatial scrolling is a bit too much. I also don't think we should have the visuals change length based on how quickly(?) you start scrolling.

I would also still rather we just measured the distance between starting point and current position, rather than projecting along that directional vector, but this is just my personal preference.

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -292,16 +306,19 @@ void HandlePointerExitAndEnter(RayEventData eventData, GameObject newEnterTarget
...
-			cachedEventData.pointerEnter = newEnterTarget;
-			t = newEnterTarget.transform;
-			while (t != null && t.gameObject != commonRoot)
+			if (!exitOnly)
 			{
-				ExecuteEvents.Execute(t.gameObject, cachedEventData, ExecuteRayEvents.rayEnterHandler);
-				if (rayEntered != null)
-					rayEntered(t.gameObject, cachedEventData);
+				// now issue the enter call up to but not including the common root
+				cachedEventData.pointerEnter = newEnterTarget;
```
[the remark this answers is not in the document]
mtschoen-unity: This has been addressed but doesn't change the line above

AndrewTHEManeri:

PR #272, opened 2017-08-16
Tiny cleanup in MathUtilsExt

mtschoen-unity:
While working on other stuff, I noticed a spelling error in the doc comment for GetTransformOffset, and also an unnecessary access to the transform property of a transform


AndrewTHEManeri:

PR #283, opened 2017-08-31
Blink Surface Detection

### LINKED ISSUE #196: Locomotion arc is doing some pretty, but odd things (Oculus)
When world is scaled, locomotion looks like: https://cl.ly/kWrg It's also nicely animated.

Very low priority but definitely a bug.

### LINKED ISSUE #209: Surface detection for blink locomotion

mtschoen-unity:
### Purpose of this PR
Fixes #209. Should also fix #196.

Pretty straightforward. Completely new BlinkVisuals stuff, and I broke out rotation from flying controls in order to do them in blink mode. Blink looked pretty bad with the unclamped alpha bug so I added the fix for that, and also fixed the ray color so it properly matches the highlight color in Linear colorspace.

I changed a bunch of IgnoreLists that were translating a list of renderers to a list of gameobjects, and back to renderers (c10ebef).

### Testing status

Tested with Oculus and Vive in PolygonAdventure scene. @dunity and @timoni also gave it a whirl.

### Technical risk

Low:  Only changes to LocomotionTool and related visuals. Other files changed due to a simple refactor to expose existing functionality via interfaces.

### Comments to reviewers

Hopefully easier to digest this time? :)

I'm not thrilled about how the actual blink point is determined by the Visuals class, but it would be stupid to do the raycasts twice or somehow try to share this functionality, in my opinion.

Something I'm curious to get feedback on is the choice of using Physics.gravity for the projectile motion equation. It cleverly avoids having to create a property for it and introduce settings and all the rest. It might get tricky, though, if users reduce gravity, since our ray shoots out at a constant speed (200m/s I think). In any case, it's a total judgment call, and easy to change


AndrewTHEManeri:
In the toolkit we use gravity as well (mostly for the direction of the arc), but the parameter we expose to the user is a 'maximum distance' parameter - an estimate of how far they want the arc to reach at the optimal (45 degree) angle.   With those two pieces of data we can figure out what speed to use for our ray to hit that distance.
I tend to prefer this more than exposing configurable speed or gravity to the users, because those tend to be 'organic' values to use - they require a lot of tweaking and testing.

mtschoen-unity:
Thanks for the suggestion @AndrewTHEManeri. Latest version will act the same regardless of gravity magnitude. Setting gravity to 0 is problematic, though

Scripts/Interfaces/FunctionalityInjection/IRaycast.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
```
dunity: #if UNITY_EDITOR

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEditor.Experimental.EditorVR.Core;
+using UnityEditor.Experimental.EditorVR.Proxies;
```
dunity: remove proxies using

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -197,20 +190,10 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
...
-				var locomotionTool = (LocomotionTool)linkedObject;
-				if (locomotionTool == this)
-					continue;
-
-				if (locomotionTool.m_State != State.Inactive)
-					return;
-			}
-
-			if (DoTwoHandedScaling(consumeControl))
+			if (DoRotating(consumeControl))
```
dunity: Why pass consumeControl to DoRotating?

mtschoen-unity: Hm. It was used, now it's not. Same as the bool return on DoCrawl, the idea is to keep these method signatures all the same in case someone comes later to refactor and needs to consume control in Rotate or changes the priority order and needs to check if crawl happened.

Down to remove the unused code. @AndrewTHEManeri for the tie-breaker?

AndrewTHEManeri: I like the symmetry, let's keep for now

AndrewTHEManeri:
Manual merge successful, perform github merge

PR #286, opened 2017-08-31
Miniworld bugfixes

### LINKED ISSUE #234: Objects Duplicate When Moving From One MiniWorld Workspace to Another (With Multiple Miniworld Workspaces Open)
**Using EVR-staging and EVR-master [0.0.7]  | Oculus + VIVE | 6/28/17**

In EVR's current state, I'm able to open multiple workspaces at one time -- and in this case, multiple MiniWorlds. This has the potential to be highly useful for a user, for example -- if within a scene with a large map, this gives me the ability to work with multiple viewpoints/areas of my scene at one time. Although this speeds up workflow (along with the features to world-scale the MiniWorld grid and two-handed scale objects), I have noticed a reoccurring issue with **moving objects from one MiniWorld to another.**

Previously, I noticed that I have the ability to drag an object outside of the MiniWorld bounds, which results in the object to appear in my physical scene (at the correct physical scale and placement I dragged it to); along with the ability to drag Assets directly from the project workspace into the MiniWorld SceneView. While there were no noticeable issues with these abilities, I wanted to test the functionality between multiple MiniWorlds. In this test, I had three MiniWorld workspaces open, each focusing on a different perspective of my scene.

Although I am able to move objects from one MiniWorld to another (the object I have selected will translate to the new MiniWorld), a copy of the selected object will appear when my object reaches the new MiniWorld bounds. (This didn't result in any issues in the Console).

**Can be seen in video below:**
https://drive.google.com/file/d/0B25RksN0ueScQmFpVHZVc3hyaDg/view?usp=sharing

**Additionally, an example in this gif:**
ezgif com-optimize

The original object remains in the scene where I placed it when moving between MiniWorlds. The duplicates of the objects can be placed throughout the scene (it seems they are legitimately just a copy, and do not affect the original object). In some instances, it appeared that the copy that appeared was the previous object selected.

**Replicate this by:**
1) Open multiple MiniWorld workspaces (2+)
2) Align the MW views to different perspectives of the scene
3) Select and drag an object from one point in a MiniWorld and drag to another MiniWorld Workspace 
4) Objects will duplicate when they are dragged out of the original MiniWord and hit the new MiniWorld workspace bounds.

mtschoen-unity:
### Purpose of this PR
Fixes #234

To reproduce (in dev):
1. Open two miniworlds
2. Move an object from one miniworld to another
3. Using the same hand, move a different object from the original MW to the second

Upon entering the second MW with the second object, the first object you manipulated will pop to the border of the second MW. This should not happen in this branch.

### Testing status

Tested with Oculus in the adventure scene.

### Technical risk

Low:  Very few changes, only in miniworlds

### Comments to reviewers

I also noticed that workspace UI was showing up in the MiniWorld, which is probably tied to the changes I made in menu-hide to allow colliders on the workspaces without blocking UI rays. The solution was to put the colliders on the parent Workspace object while putting the WorkspaceUI object back into the UI layer (thus culled by the miniworld camera). This had the added benefit of using IWorkspace as the component type we use in IMenu, decoupling the menu implementation from WorkspaceUI.

### DIFF
--- a/Scripts/Core/EditorVR.Menus.cs
+++ b/Scripts/Core/EditorVR.Menus.cs
@@ -42,7 +42,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
 
 			// Local method use only -- created here to reduce garbage collection
 			readonly List<DeviceData> m_ActiveDeviceData = new List<DeviceData>();
-			readonly List<WorkspaceUI> m_WorkspaceComponents = new List<WorkspaceUI>();
+			readonly List<IWorkspace> m_WorkspaceComponents = new List<IWorkspace>();
 			readonly Collider[] m_WorkspaceOverlaps = new Collider[k_PossibleOverlaps];
 
 			public Menus()
--- a/Scripts/Core/EditorVR.MiniWorlds.cs
+++ b/Scripts/Core/EditorVR.MiniWorlds.cs
@@ -123,6 +123,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
 					{
 						hash.RemoveFromSpatialHash(grabData.transform.gameObject);
 						grabData.SetScale(scaleFactor);
+						grabData.Update(originalRayOrigin);
 					}
 				}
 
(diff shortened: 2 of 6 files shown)


AndrewTHEManeri:

PR #291, opened 2017-09-06
Fix inspector workspace

mtschoen-unity:
@amydigiov noticed a regression that slipped through with my menu-hide branch. The change to workspace minbounds caused the size property of the inspector listview to never be set, leaving it with a size of 0. This caused all rows to show up hidden, even when the inspector was populated with data.

The fix is just to kick the bounds at start, like other workspaces that require size to be set at start (Hierarchy, etc.). I also noticed that the default bounds still had a height of 40cm which isn't sensible.


AndrewTHEManeri:
Conflict resolution required and applied, performing github merge

PR #297, opened 2017-09-15
Prevent interaction with the standard manipulator by RayOrigins whose cones/rays are disabled

### LINKED ISSUE #296: Standard manipulator can be remotely controlled (via ray) when performing grab locomotion
The standard manipulator handles can be remotely controlled (via ray) when in the midst of performing a grab locomotion action.  While the grip button is held, you can aim at an active standard manipulator handle, then use trigger-1 to manipulate the handle/selected-object.  This prevents the triggering of spatial selection (hold grip, and press trigger-1) in this state.

dunity:
**Purpose of this PR**

Resolves #296 

Prevent interaction with the standard manipulator by rayOrigins whose cones or rays are disabled (during grab locomotion, etc).  This was preventing the display of spatial-selection in such cases as well.

Verify that ray and cone are visible before adding rayOrigin to drag or hover sources.

I could have implemented this in BaseManipulator (by checking all handles via a property, or similar), but felt it would have fewer potential repercussions if implemented at a lower level, in the BaseHandle itself.  This also skipped the need to add any additional logic to the TransformTool's ProcessInput() function.

**Testing status**

Tested in development; found no issues.

**Technical risk**

Low: I haven't seen any side effects from this.

**Comments to reviewers**

Is there a case in which we would want to allow a rayOrigin to be added to either collection, if either was hidden?


Scripts/Handles/BaseHandle.cs
```diff
@@ -106,6 +106,9 @@ public void OnBeginDrag(RayEventData eventData)
 				return;
 
 			var rayOrigin = eventData.rayOrigin;
+			if (!this.IsConeVisible(rayOrigin) || !this.IsRayVisible(rayOrigin))
+				return;
```
mtschoen-unity: Why do this in BaseHandle, not MultipleRayInputModule?

dunity:
Addressed input and went with arch suggestion made in #298.  All locomotion modes are handled now as well.

Scripts/Handles/BaseHandle.cs
dunity: Added IBlockUIInteraction check in AddRaycastSource valid check (in EditorVR.Rays).

AndrewTHEManeri:

PR #301, opened 2017-09-21
Basic proxy feedback

mtschoen-unity:
### Purpose of this PR

Set up a system for controlling visual feedback on the controllers (and potentially elsewhere)

### Testing status

Automated tests all green. The is not very thoroughly tested, to be honest, and LocomotionTool has a lot of permutations for what button does what and how to transition in and out of those states. Do try it out and let me know if you find any bugs or annoyances. We intend to iterate on this, but it will likely stick around for a while.

### Technical risk

Low:  Mostly adds new functionality, and it's pretty light on code. Worst case scenario we get some annoying tooltips, but they auto-hide after 5 seconds anyway.

### Comments to reviewers

I had to update the TooltipModule and HighlightModule to support disabling after a delay. This could have been done with coroutines in TwoHandedProxyBase, but it seemed like an appropriate system-wide feature.

Speaking of TwoHandedProxyBase, I would still like to re-work how Proxies work (maybe as my next controller-related task). Basically I would like to remove the concept of two hands and turn this class into just ProxyBase, which allows some kind of set up for how many nodes get RayOrigins and the like. What I would like to do is make it possible to integrate tracker pucks and sensor poses so that people can optionally make use of these features. Another improvement I would like to make is removing the disabled proxies from the scene and actively creating/destroying proxies when inputs become active.


Scripts/Utilities/InputUtils.cs
```diff
@@ -42,6 +42,31 @@ public static void CollectDeviceSlotsFromActionMapInput(ActionMapInput actionMap
 				deviceSlots.Add(deviceSlot);
 			}
 		}
+
+		public static void GetBindingDictionaryFromActionMap(ActionMap actionMap, Dictionary<string, List<VRInputDevice.VRControl>> bindingDictionary)
```
AndrewTHEManeri: Nice

Tools/VacuumTool/VacuumTool.cs
```diff
@@ -23,38 +28,77 @@ sealed class VacuumTool : MonoBehaviour, ITool, ICustomActionMap, IUsesRayOrigin
...
+					if (m_Feedback.Count == 0)
+					{
+						foreach (var kvp in m_Controls)
+						{
+							foreach (var id in kvp.Value)
+							{
+								var request = new ProxyFeedbackRequest
+								{
+									control = id,
+									node = node.Value,
```
AndrewTHEManeri: Are workspaces the only vacuumables?

Scripts/Modules/TooltipModule/TooltipModule.cs
```diff
@@ -91,21 +94,28 @@ void Update()
 
 				if (!IsValidTooltip(tooltip))
 					m_TooltipsToHide.Add(tooltip);
+
+				if (tooltipData.persistent)
```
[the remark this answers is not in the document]
mtschoen-unity: Not quite the same manner. Persistent tooltips will only be hidden if duration > 0. Non-persistent tooltips will be hidden as soon as a RayExit occurs, because HideTooltip is called without the persistent flag. The whole point of a persistent tooltip is that it won't be hidden by ray exits. Similarly, we |= persistent, and only set its duration if >0 so that Ray Enters don't unset persistent states.

Scripts/Modules/TooltipModule/TooltipModule.cs
```diff
@@ -91,21 +94,28 @@ void Update()
...
 		static Transform GetTooltipTarget(ITooltip tooltip)
 		{
 			var placement = tooltip as ITooltipPlacement;
-			var target = ((MonoBehaviour)tooltip).transform;
 			if (placement != null)
-				target = placement.tooltipTarget;
-			return target;
+				return placement.tooltipTarget;
+
+			return ((MonoBehaviour)tooltip).transform;
```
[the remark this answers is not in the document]
mtschoen-unity: Yes. This is why I made this change in the first place, because I ended up trying a non-MB tooltip. But in this case you would never be able to use something that was just ITooltip and not a monobehaviour. I'd want some sort of error in this case, rather than the system silently doing nothing.

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -161,6 +182,100 @@ public virtual void Update()
...
+						if (matchButton && priority && (matchChanged || sameCaller))
+							request = req;
+					}
+
+					if (request == null)
+						continue;
+
+					var button = kvp.Value;
+					if (button.renderer)
+						this.SetHighlight(button.renderer.gameObject, true, duration: k_FeedbackDuration);
```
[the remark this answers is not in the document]
mtschoen-unity: Yep. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -161,6 +182,100 @@ public virtual void Update()
...
+		void RemoveFeedbackRequest(ProxyFeedbackRequest request)
+		{
+			Dictionary<VRInputDevice.VRControl, ProxyHelper.ButtonObject> buttons;
+			if (m_Buttons.TryGetValue(request.node, out buttons))
+			{
+				ProxyHelper.ButtonObject button;
+				if (buttons.TryGetValue(request.control, out button))
+				{
+					if (button.renderer)
+						this.SetHighlight(button.renderer.gameObject, false);
```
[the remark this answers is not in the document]
mtschoen-unity: Yes. This will unset highlights/tooltips on this button from any source. This is why we call ExecuteFeedback afterward, which will re-execute any underlying feedback which we might have stomped.

Scripts/Utilities/InputUtils.cs
mtschoen-unity: Heh, yeah it's ugly but it works. I still don't like that I need to hardcode action names, but I couldn't think of a better system. Also, I bet this (and the code for locking/consuming controls) has to change when we update to the new system.

Tools/VacuumTool/VacuumTool.cs
mtschoen-unity: Currently, yes. I figured the UI hint saying "vacuumable" would be confusing.

Scripts/Modules/TooltipModule/TooltipModule.cs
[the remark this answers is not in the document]
AndrewTHEManeri: The follow up question is why highlights don't need to follow this pattern as well?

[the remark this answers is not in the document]
mtschoen-unity: They aren't triggered on hover for anything but scene objects. Maybe they should show up on hover, for the buttons but I didn't think that was necessary at this stage

Scripts/Utilities/IntersectionUtils.cs
```diff
@@ -185,7 +186,10 @@ public static void SetupCollisionTester(MeshCollider collisionTester, Transform
 		{
 			var mf = obj.GetComponent<MeshFilter>();
 			if (mf)
+			{
 				collisionTester.sharedMesh = mf.sharedMesh;
+				collisionTester.transform.localScale = Vector3.one;
+			}
 
 			if (!mf)
```
dunity: Not that it's in this PR, but why not have this be an else, to skip the additional test if the MeshFilter isn't null?

Tools/VacuumTool/VacuumTool.cs
dunity: Null check node before retrieving/assigning node.Value.  Looks like nulls can't occur in the current arch, but wanted to mention it.

Scripts/Interfaces/Capability/IFeedbackReciever.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace UnityEditor.Experimental.EditorVR
+{
+	/// <summary>
+	/// Implementors can receive Feedback Requests
+	/// </summary>
+	public interface IFeedbackReciever
+	{
+		void AddFeedbackRequest(FeedbackRequest request);
+		void RemoveFeedbackRequest(FeedbackRequest request);
+		void ClearFeedbackRequests(IRequestFeedback caller);
```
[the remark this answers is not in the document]
AndrewTHEManeri: But the function clears all feedback requests if no caller is supplied.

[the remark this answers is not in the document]
mtschoen-unity: Hm. This is an interesting one. It's actually set up so that you can't call it without a caller bc the extension method always sends along the caller. The case where we call it with null can only happen from the module.

Thanks for bringing this up, because all three of these methods should probably have docs, even though it's an internal interface (for now).

Tools/VacuumTool/VacuumTool.cs
mtschoen-unity: Yeah. My other thought was just to make the field a Node? so it can just be passed directly through. Otherwise that's a lot of null checks (here and in all the ShowXRequest methods in LocomotionTool).

@AndrewTHEManeri thoughts?

AndrewTHEManeri: Is there expected behavior for objects if they encounter a null node?  And if a 'null' value is expected, should the node enum itself just get an extra value?

mtschoen-unity: Yeah, a null node is meant to indicate that there is no hand/node involved. I'm also in favor of a "None" value for Node, and getting rid of all the Node?s

dunity: I'm for "None", in order to lose the Null.

dunity:
It looks like SelectionTool & the LocomotionTool are adding more requests than they're removing in TwoHandedProxyBase.  Though the VacuumTool is removing as many as it is adding.

AndrewTHEManeri:

PR #323, opened 2017-09-28
fix blink locomotion when changing player scale

### LINKED ISSUE #300: Blink visuals don't hide when triggered before changing ViewerScale
The blink visuals stay visible (after the blink button is released), and don't perform a blink when the following action is performed:

1: Press & hold the blink button
2: While keeping the blink button held, trigger a two-handed viewerScale action.
3: While performing the viewerScale change, release the held blink button.
4: End the viewerScale action (release both grips).
5: The blink visuals stay visible, and perform no blink action in this case.

stella3d:
fixes #300 

tested that everything else in locomotion still works as expected.  
now, if you release the blink button while in the middle of changing player scale, the blink is cancelled.

I'm _not wild_ about having to do it like this - with the `if(DoThing) return;` pattern, but as far as the scope of this bugfix goes, i think it's the right solution.
Perhaps there is a better way ?  it seems like this pattern in general will lead to more instances of control being consumed in a way that prevents another system from working.

### DIFF
--- a/Tools/LocomotionTool/LocomotionTool.cs
+++ b/Tools/LocomotionTool/LocomotionTool.cs
@@ -199,7 +199,12 @@ namespace UnityEditor.Experimental.EditorVR.Tools
 			this.SetUIBlockedForRayOrigin(rayOrigin, true);
 
 			if (DoTwoHandedScaling(consumeControl))
+			{
+				if (m_Preferences.blinkMode)
+					if (DoBlink(consumeControl))
+						return;
 				return;
+			}
 
 			if (DoRotating(consumeControl))
 				return;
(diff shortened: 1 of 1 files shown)


AndrewTHEManeri:
I think the 'proper' solution to this is instead of returning when a locomotion method is used, recording that fact and passing it down to each function in the line, so that they can handle the situation themselves.  In that scenario, activating two-handed scale could just cancel the blink out.  I'll make an issue for this.

But, in terms of a fix for this release I'm fine with this solution

stella3d:
ok, changing it to cancel blink

stella3d:
blink cancels now !

mtschoen-unity:
Yeah... the if (didlocomote) return pattern was a little more sensible when there wasn't anything that needed to happen when transitioning between states. Blink visuals and feedback really complicated things.

I'm open to restructuring this tool, but I'm not sure of the best solution. Maybe something more like a state machine?

stella3d:
@mtschoen-unity something like that is probably sensible, but that's a bigger discussion.  for the purposes of this release i think just fixing this one bug instance is what we want

AndrewTHEManeri:
Excellent

PR #328, opened 2017-09-28
Don't spatial-select tools until input is released

### LINKED ISSUE #322: Quickly spatial selecting while over a transform manipulator can break the ability to select things
To Repro - have a few tools open so that spatial selecting can occur.  Choose the selection tool.  Select an object to make the transform manipulator appear.  While the ray is over one of the handles, activate a spatial select as quickly as possible.  If done (in)correctly, most of the manipulators will dissappear - but not all of them - and the ability to select anything is gone.

Sometimes (but not always) this will trigger an error "Gameobject is already being activated or deactivated"  for the handle being manipulated - during the Transform tool's OnSelectionChanged function

mtschoen-unity:
The fact that you could switch tools with the trigger held was causing all sorts of chaos.  There are probably clever input system hacks to prevent this issue. ConsumeControl doesn't work in this case because we are messing with actionmapinputs while changing tools. The input fixes I made yesterday are still valid.

Fixes #322.

### DIFF
--- a/Menus/ToolsMenu/ToolsMenu.cs
+++ b/Menus/ToolsMenu/ToolsMenu.cs
@@ -200,7 +200,7 @@ namespace UnityEditor.Experimental.EditorVR.Menus
 					m_ToolsMenuUI.HighlightSingleButtonWithoutMenu((int)(buttonCount * normalizedRepeatingPosition) + 1);
 				}
 			}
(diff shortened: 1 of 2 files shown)


AndrewTHEManeri:
This does make the issue less likely to occur but I have a more consistent repro case for 322

PR #346, opened 2017-10-03
Proxy Animator

mtschoen-unity:
### Purpose of this PR

Add a component which maps controller input to articulation of the controller model (again, Touch-only)

### Testing status

I've tested this on Touch with OVR integration (i.e. not via SteamVR). We decided that I should go through with the IProcessInput refactor, which means that this can potentially break anything input-related. I have not done a thorough test of all systems. If we are to do a Functionality Injection refactor soon, it makes sense to also change how modules are loaded and initialized. Throughout this process, similar instabilities can be introduced.

### Technical risk

High:  Input processing and interface connecting works differently now. I did not ensure the processing order of workspaces, menus, etc. on purpose, because I wanted to see if explicit ordering of internal systems was necessary. It appears not to be necessary. Process order on tools still happens in EditorVR in stack order.

### Comments to reviewers

The implementation of ProxyAnimator is intentionally simple. Dylan made a good call about letting proxy creators just set axes and limits. It works remarkably well and I think this covers the 90%. The 10% are welcome to implement their own ProxyAnimator equivalent. The setup is currently hardcoded into OnProxyActiveChanged, because we that is the most convenient point to connect its interfaces. Due to all of the changes in the works, I didn't want to put too much thought into how to solve this problem more generally yet.

I implemented "ignoreLocking" after the priority order feature. I incorrectly assumed that processing the animator first in the queue would avoid the need to ignore locking, but because the actionmap itself gets the reset calls, any locked control will simply not animate unless we avoid the reset. Thus, putting ProxyAnimator first in the order (and the order feature itself) is unnecessary. That said, I think it makes sense that they be processed first to demonstrate  the feature, ensure it still works, and make it clear they're separate from other input processors. I also considered just removing the order feature, but I'm pretty sure we'll need it in the future. I'm not sure if we should prevent tools from using ignoreLocking, but now that it exists, it would be possible for tools to conflict. Also, the reason why ignoreLocking is done in ICustomActionMap is that it has to be associated with the created actionMapInput. It would be impossible for different implementors of IStandardActionMap (at least as we have implemented it, with it sharing an AMI) to ignore or not ignore locking. The important thing to remember is that different users of a single AMI instance can't disagree about ignoring locking.

To explain the reason I refactored IConnectInterfaces:  I needed to allow ProxyAnimators to be connected to an inputDevice, and uncovered some ugliness.  Since the IProcessInput refactor had wide-ranging changes to begin with, I decided to fix this now as well. We had an internal overload of ConnectInterface that took an inputDevice and just translated it into a rayOrigin. Now everything accesses ConnectInterfaces through the interface, and I've provided a method for translating InputDevices to RayOrigins where needed. This has some upsides, but it's not perfect (yet). Basically everything we've been doing so far could connect on just the object and sometimes a ray origin. This alternate DeviceInput version of ConnectInterfaces lead me to think we might want to supply a different payload. The problem is that ConnectInterfaces is a one-shot process. If one module needs one type of payload, and another needs a different one, we're stuck because we can only call ConnectInterfaces on the instance once. We could make userData an array, or come up with a way of subclassing payloads to get multiple payloads out of one object. It maybe be possible to use a params object[] signature to basically implement whatever combinations of payloads we want as overloads.

In any case, I stopped here with just generalizing the rayOrigin field to an object called userData. In this case, we could more easily support different payload types in silos. Essentially as long as you don't need a rayOrigin you can send whatever other type of payload you want. The ConnectInterfaces refactor is contained to 82366db and 8c8587f.

I considered stripping this into its own PR, but the ProcessInput refactor went pretty far at the outset, and otherwise the ProxyAnimator changes are very minimal. The only thing specific to ProxyAnimator here is the class itself and the additions to ProxyHelper.ButtonObjects.


amirebrahimi:
I don't usually comment, however, I noticed that this PR is moving interface connection back into the modules. I'd suggest keeping the module connectors to separately connect interfaces as it keeps the coupling loose. It means modules can go to other projects without having to bring in the interfaces. The module connector is what binds those two. Furthermore, it leaves freedom to map the interfaces to other implementations outside of the module. This would be one of those situations where a class is making changes to other peer classes/interfaces instead of down.

mtschoen-unity:
Thanks for the comment. I forgot the purpose of the connector classes, and thought they were vestigial. The main motivation was to expose ConnectInterfaces to modules via functionality injection. It does make sense that IInterfaceConnector should stay out of the modules, though. I'll revert those changes.

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -6,15 +6,14 @@
...
-using UnityEngine.InputNew;
 
 namespace UnityEditor.Experimental.EditorVR.Core
 {
 	partial class EditorVR
 	{
 		const float k_MainMenuAutoHideDelay = 0.125f;
 		const float k_MainMenuAutoShowDelay = 0.25f;
-		class Menus : Nested, IInterfaceConnector, ILateBindInterfaceMethods<Tools>
+		class Menus : Nested, IInterfaceConnector, ILateBindInterfaceMethods<Tools>, IConnectInterfaces
```
AndrewTHEManeri: Behavior of this seems slightly different now - previously you could add/remove a mainMenu with a null rayorigin - now code stops that from happening.  Was anything relying on that behavior before?

Scripts/Core/EditorVR.DirectSelection.cs
```diff
@@ -35,9 +35,9 @@ public DirectSelection()
 				IGetPointerLengthMethods.getPointerLength = GetPointerLength;
 			}
 
-			public void ConnectInterface(object obj, Transform rayOrigin = null)
+			public void ConnectInterface(object @object, object userData = null)
```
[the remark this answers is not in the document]
mtschoen-unity: No technical reason. Our conventions instruct us to "avoid abbreviations where possible," so I figured while I was changing all of the uses of ConnectInterfaces, I might as well try this out.

I'm ambivalent about it, and happy to go back to "obj" if we prefer that.

Scripts/Core/EditorVR.Menus.cs
mtschoen-unity: I'm not sure this is any different. You can't add null keys to dictionaries. We should probably have been doing that null check beforehand. In fact, I'm not 100% sure it's required now, since we only call ConnectInterfaces on menus with ray origins.

I was careful to look for all usages of ConnectInterfaces to ensure that callers that didn't provide a ray origin weren't affected.

Scripts/Core/EditorVR.DirectSelection.cs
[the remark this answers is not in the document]
AndrewTHEManeri: I prefer something without the @ - interface connecting is one of the spookier parts of our codebase, anything we can do to make it less exotic to the user, the better.

Scripts/Core/EditorVR.Menus.cs
AndrewTHEManeri: Sounds good

Scripts/Core/EditorVR.DirectSelection.cs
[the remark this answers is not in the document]
mtschoen-unity: That's reasonable. So back to obj? Maybe "target"? connectee?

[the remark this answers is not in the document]
AndrewTHEManeri: I think target works good here - especially since the second parameter is also kind of 'generic' now

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -0,0 +1,136 @@
...
+		m_Buttons = GetComponent<ProxyHelper>().buttons;
+	}
+
+	public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
+	{
+		if (m_Buttons == null)
+			return;
+
+		var length = m_Buttons.Length;
+		if (m_Controls == null)
```
AndrewTHEManeri: Is there some kind of order-of-operations thing where this has to happen in ProcessInput?
Should there be some kind of higher level "InitializeInput" function that IProcessInput implementors need to handle?

Scripts/Core/EditorVR.DirectSelection.cs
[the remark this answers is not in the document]
mtschoen-unity: Yeah. My thought process here was to try and generalize ConnectInterfaces. Once I finish the tabs->spaces update I'll change to "target" everywhere

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -0,0 +1,136 @@
...
+			var control = m_Controls[i];
+			var info = m_TransformInfos[button.transform];
+
+			//Assume control values are [-1, 1]
+			var min = button.min;
+			var offset = min + (control.rawValue + 1) * (button.max - min) * 0.5f;
+
+			var positionOffset = info.positionOffset;
+			var translateAxes = button.translateAxes;
+			if ((translateAxes & AxisFlags.X) != 0)
```
AndrewTHEManeri: you might consider an extension method that converts AxisConstraints to a direction vector

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -161,6 +182,100 @@ public virtual void Update()
...
+					}
+
+					if (request == null)
+						continue;
+
+					var button = kvp.Value;
+					if (button.renderer)
+						this.SetHighlight(button.renderer.gameObject, true, duration: k_FeedbackDuration);
+
+					if (button.transform)
```
AndrewTHEManeri: What exactly is a button without a transform?

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -161,6 +182,100 @@ public virtual void Update()
...
+		{
+			var proxyRequest = request as ProxyFeedbackRequest;
+			if (proxyRequest != null)
+			{
+				m_FeedbackRequests.Add(proxyRequest);
+				ExecuteFeedback(proxyRequest);
+			}
+		}
+
+		void ExecuteFeedback(ProxyFeedbackRequest changedRequest)
```
AndrewTHEManeri: Function seems to be utilized in a weird way?  When called through AddFeedbackRequest, it should always match matchChanged and sameCaller and thus continue on.

When called through RemoveFeedbackRequest, it will always fail the matchChanged and sameCaller request and essentially do nothing.

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -35,14 +35,18 @@ sealed class SelectionTool : MonoBehaviour, ITool, IUsesRayOrigin, IUsesRaycastR
 		Color m_MultiselectRayColor;
 		bool m_MultiSelect;
 
+		readonly Dictionary<string, List<VRInputDevice.VRControl>> m_Controls = new Dictionary<string, List<VRInputDevice.VRControl>>();
```
AndrewTHEManeri: we might want to consider some kind of container class for this - kind of a big function signature to drop on everyone

Scripts/Proxies/ProxyAnimator.cs
mtschoen-unity: This isn't a bad idea. The reason it has to happen here is because this is (by design) the only place where we have access to the ActionMapInput which was created from our ActionMap. The idea was to make it awkward for people to use input in an Update method because then we can't control the order in which input is processed.

That said, there are many other places where we break this convention, for example ILinkedObject tools that create a private field to store the ActionMapInput so that the shared updater can access input from linked tools. They would also benefit from returning to the original implementation of ICustomActionMap, which included an ActionMapInput property with a public setter that the system would use to give back the AMI that it creates when it spawns the tool.

Scripts/Proxies/ProxyAnimator.cs
mtschoen-unity: Great idea. I'm not exactly sure where it should go. Should I add a static class to AxisFlags.cs? Or maybe in MathUtils?

Tools/SelectionTool/SelectionTool.cs
mtschoen-unity: Hm. So it would just be a container that wraps the big fat dictionary type? I guess that's not a bad idea, especially since we could use the same kind of thing to cut down on that really really long string of generics in TwoHandedProxyBase

Scripts/Proxies/TwoHandedProxyBase.cs
mtschoen-unity: It could be one that we only highlight and has no tooltip. It would also be a shame to throw null refs everywhere if an asset is deleted or for some other reason the transform reference goes missing. Should we write a warning to the console in an else block?

Scripts/Proxies/TwoHandedProxyBase.cs
mtschoen-unity: It won't always fail matchChanged. We match on everything but the caller. This happens specifically because of the case where you remove feedback that was overriding an existing piece of feedback. This happens in the case of vacuum tool overriding the "crawl" feedback on the grip and then removing itself. When you remove the vacuum feedback, the crawl feedback matches and executes.

I ended up fixing a bug with this logic during my second Feedback pass, if that version makes more sense... https://github.com/Unity-Technologies/EditorVR/commit/4cae9b52ca01ba6bcf9cc4d8d87352d69ad57c97#diff-772d21b9cb82f8b9c5868499ff7446e9R220

mtschoen-unity:
@AndrewTHEManeri I ended up going a different way to deal with those big generic dictionaries. I used a type alias. I'm not sure how that shows up in intellisense, but it does make working with those types a lot easier. Does that work for you?

AndrewTHEManeri:
Yeah, that all looks good to me.

AndrewTHEManeri:

PR #347, opened 2017-10-04
EVR development on Mainline Unity

AndrewTHEManeri:
This PR moves over the initial set of changes needed to run EditorVR in 2017.2.

Already tested with a manual local merge of development - no blocking errors.


Tests/Editor/CompilationTest.cs
```diff
@@ -30,7 +30,7 @@ public void NoCCUDefines()
 		public void NoEditorVR()
 		{
 			var defines = EditorUserBuildSettings.activeScriptCompilationDefines.ToList();
-			defines.Remove("UNITY_EDITORVR");
+			defines.Remove("UNITY_2017_2_OR_NEWER");
```
mtschoen-unity: I wasn't sure at first whether this was doing the job, but it is. The defines list includes the entire list of defines.

That said, I have found some issues in other Unity versions, and am working on a commit to sort them out.

Tools/AnnotationTool/AnnotationTool.cs
```diff
@@ -313,7 +313,7 @@ void HandleBrushSize(float value)
 		if (m_AnnotationPointer != null)
 		{
 			var brushSize = m_Preferences.brushSize;
-			if (VRSettings.loadedDeviceName == "OpenVR") // For vive controllers, use 1:1 touchpad setting.
+			if (UnityEngine.XR.XRSettings.loadedDeviceName == "OpenVR") // For vive controllers, use 1:1 touchpad setting.
```
mtschoen-unity: This should have been fixed before AnnotationTool landed (my bad) but I'm noticing it here.

This should use IUsesProxyType, rather than checking the device name.

Scripts/Utilities/ObjectUtils.cs
```diff
@@ -278,9 +278,26 @@ public static string NicifySerializedPropertyType(string type)
...
+			foreach (var assembly in assemblies)
+			{
+				try
+				{
+					foreach (var type in assembly.GetTypes())
+					{
+						if (type.Name.Equals(name) && typeof(UnityObject).IsAssignableFrom(type))
+						{
+							return type;
+						}
```
mtschoen-unity: remove brackets on single line if

Scripts/Core/VRView.cs
```diff
@@ -151,7 +151,8 @@ public void OnEnable()
 			// VRSettings.enabled latches the reference pose for the current camera
 			var currentCamera = Camera.current;
 			Camera.SetupCurrent(m_Camera);
-			VRSettings.enabled = true;
+			UnityEngine.XR.XRSettings.enabled = true;
+			UnityEngine.XR.InputTracking.Recenter();
```
mtschoen-unity: This crept back in. Do we need it for 2017.2 now?

For clarity, this was removed to fix the "workspaces are at a weird rotation" bug, #197

Scripts/Core/EditorVR.cs
```diff
@@ -299,7 +299,7 @@ static void ClearDeveloperConsoleIfNecessary()
 
 				if (result)
 				{
-					var logEntries = asm.GetType("UnityEditorInternal.LogEntries");
+					var logEntries = asm.GetType("UnityEditor.LogEntries");
 					var clearMethod = logEntries.GetMethod("Clear", BindingFlags.Static | BindingFlags.Public);
 					clearMethod.Invoke(null, null);
```
mtschoen-unity: Is there still no API for this?

Scripts/Core/VRView.cs
AndrewTHEManeri: No, it can go

Scripts/Utilities/ObjectUtils.cs
AndrewTHEManeri: Not quite sure why the inner part was de-linqified

amirebrahimi: Because exception handling is now needed

amirebrahimi:
Unless I've missed something - if you remove all of the UNITY_2017_2_OR_NEWER, we'll lose the graceful failure that we had that pointed users to upgrade.

AndrewTHEManeri:
From our discussion - this will gracefully fallback on 2017.1. 5.6 or earlier the namespaces have changed.

I'm alright with this - the release notes are pretty straightforward about what versions of Unity you can use this with.

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -109,6 +109,7 @@ static void ReopenOnExitPlaymode()
 			}
 		}
 
+#if UNITY_2017_2_OR_NEWER
```
AndrewTHEManeri: Rather than having the Unity_2017_2_OR_NEWER check for each of these, let's have one that wraps all three functions, and an else that defines empty versions of them

Scripts/Core/VRView.cs
```diff
@@ -333,6 +346,7 @@ void UpdateHMDStatus()
 
 		static bool GetIsUserPresent()
 		{
+#if UNITY_2017_2_OR_NEWER
```
AndrewTHEManeri: Kind of want to do a bigger define wrapper here as well but I see things aren't quite as adjacent/easily grabbed together.  Thoughts?

amirebrahimi: Well, there isn't a need for backwards compatibility moving forward, so why should anything compile from EditorVR when the version of Unity is not 2017.2 or newer?

AndrewTHEManeri: The danger with blocking out *all* of the EVR code behind this kind of fence is that a user can hose their scriptable objects (like their editing contexts) because their meta files and references are just meaningless now.  This has happened to us more than once already during this upgrade.

Scripts/Core/EditorVR.cs
AndrewTHEManeri: Not that I have found

Scripts/Core/VRView.cs
mtschoen-unity: It also has to return *something* or it will cause compile errors.

mtschoen-unity:
Re: 2017_2_OR_NEWER stuff: yes, I was using 5.5 (was the earliest vanilla version I had installed) as a test bench to hunt down compile error when removing 2017-specific stuff. My intention was to, within reason, wrap only what needed to be compiled out in order to get rid of all warning/errors in 5.5, then go back to 2017 and make sure nothing got hit by the API updater.

Scripts/Core/VRView.cs
mtschoen-unity: I also didn't want to have to refactor uses of methods, so as much as possible I just focused on implementations.

Scripts/Core/Contexts/EditingContextManager.cs
mtschoen-unity: We wouldn't need empty versions, since OnEnable and OnDisable aren't explicitly called. I can just do one big one with no else.

AndrewTHEManeri:

PR #348, opened 2017-10-06
Chipping away at our per-frame allocations. 

AndrewTHEManeri:
### Purpose of this PR
Part 1 of a series of weekly adventures in to reducing EditorVR's allocations-per-frame.  This one takes things down by about 1-3kb.  In the grand scheme of things, that's about 5% of our blank-scene allocations-per-frame.

### Testing status

Tested in Adventure test scene and blank scene, both with the profiler open.  Actual functions changed were measured to ensure their contribution to allocations had gone down.

### Technical risk

Medium - many of these were hand-rewriting convenience functions which means that human error can now be introduced into lookup code that happens very frequently.  Biggest 

### Comments to reviewers

I'm targeting the old .net version which means that some of these optimizations might automatically occur with the experimental .net version.  Since we can't guarantee our users are using that (or that Input is stable with it), I have to optimize for the lowest common denominator.

Much of the allocations are happening on the editor side, but EVR itself is contributing to that too, so my goal is to remove us from that situation.  This looks to be a 'death by 1000 papercuts' type deal, so it will be ongoing process to take this as low as we can.


Scripts/Core/EditorVR.Menus.cs
```diff
@@ -382,7 +382,21 @@ internal static bool IsValidHover(MultipleRayInputModule.RaycastSource source)
...
+				while (dataCounter < deviceDataList.Count)
+				{
+					var currentDevice = deviceDataList[dataCounter];
+					if (currentDevice.rayOrigin == rayOrigin)
+					{
+						deviceData = currentDevice;
+						break;
+					}
+					dataCounter++;
+				}
```
mtschoen-unity: Why not just a for/foreach loop here? AFAIK, as of 5.6 foreach on lists and dictionaries no longer allocate. I think it was just the use of LINQ that got us here.

Scripts/Core/EditorVR.Viewer.cs
```diff
@@ -37,6 +37,8 @@ class Preferences
 
 			const float k_CameraRigTransitionTime = 0.25f;
 
+            static Collider[] s_CachedColliders = new Collider[20];
```
mtschoen-unity: Const the "20." Also, looking at the context here, I don't think more than 2 are necessary. We're specifically looking for the player body triggers, one directly above the head and one slightly behind and above the head.

Scripts/Modules/DeviceInputModule.cs
```diff
@@ -37,6 +37,7 @@ sealed class DeviceInputModule : MonoBehaviour
 
 		public Action<HashSet<IProcessInput>, ConsumeControlDelegate> processInput;
 		public Action<List<ActionMapInput>>  updatePlayerHandleMaps;
+		List<InputControl> m_RemoveList = new List<InputControl>();
```
mtschoen-unity: We've been commenting these types of lists specifically with this:

Any of these lists can (and should?) be static readonly, because the idea is to clear them out before each use, and they aren't being used in methods called concurrently. This will reduce the size of the object and I think it also avoids the offset when getting the reference.

Scripts/Utilities/ComponentUtils.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+	/// <summary>
+	/// Special utility class for getting components in the editor without allocations
+	/// </summary>
+	/// <typeparam name="T"></typeparam>
+	public class ComponentUtils<T>
+	{
+		static List<T> s_RetrievalList = new List<T>();
+
+		public static T GetComponent(GameObject go)
```
mtschoen-unity: Why not an extension method? and a static class?

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -39,6 +39,8 @@ sealed class SelectionTool : MonoBehaviour, ITool, IUsesRayOrigin, IUsesRaycastR
 
 		readonly Dictionary<Transform, GameObject> m_SelectionHoverGameObjects = new Dictionary<Transform, GameObject>();
 
+        Dictionary<Transform, GameObject> m_TempHovers = new Dictionary<Transform, GameObject>();
```
mtschoen-unity: Again, comment and static readonly here

Scripts/Core/EditorVR.Viewer.cs
mtschoen-unity: Actually, never mind. If there are a bunch of colliders near your head, that would fail. Something in me still wants this to be a power of two...

Scripts/Utilities/ComponentUtils.cs
mtschoen-unity: Actually, maybe you can't do the static class because of the templting. And a static method would no doubt need a different name. I can check and see if it's possible.

AndrewTHEManeri: Yes - this class and method has to stay because of the templating.  We could wrap it in an extension method (as we are explicitly calling out what object/interface we are using), but I'm not convinced we want to.  Unity already has methods for getting a component from a gameobject and I don't want to confuse users by adding duplicate methods on their intellisense - especially since this is for an EVR specific optimization.

Additionally, our ObjectUtils.AddComponent functions are also not extension methods.

Scripts/Core/EditorVR.Viewer.cs
AndrewTHEManeri: I'm down with 32

Scripts/Core/EditorVR.Menus.cs
AndrewTHEManeri: That sounds right - I'll test & confirm

Scripts/Utilities/ComponentUtils.cs
mtschoen-unity: Just tried the extension method, and indeed we can't put extension methods in templated classes. However, we can use a templated static class, which makes it clear that you should never instantiate ComponentUtils. I would still say we should make ComponentUtils static.

Scripts/Modules/DeviceInputModule.cs
AndrewTHEManeri: That's why I put it in this section where the comment was -I'll make it static

mtschoen-unity: Oh hmm. This one is a little weird. The public properties aren't gc related

AndrewTHEManeri: I'll move above the puplic props

Scripts/Core/EditorVR.Menus.cs
AndrewTHEManeri: correct - re-foreach'd

Scripts/Core/EditorVR.UI.cs
```diff
@@ -129,7 +129,16 @@ void SetManipulatorsVisible(ISetManipulatorsVisible setter, bool visible)
 
 			bool GetManipulatorDragState()
 			{
-				return m_ManipulatorControllers.Any(controller => controller.manipulatorDragging);
+			    var controllerCounter = 0;
+			    while (controllerCounter < m_ManipulatorControllers.Count)
```
mtschoen-unity: Can also be a foreach

Scripts/Core/EditorVR.Viewer.cs
```diff
@@ -199,11 +203,15 @@ static bool Overlaps(Transform rayOrigin, Collider trigger)
 			{
 				var radius = DirectSelection.GetPointerLength(rayOrigin);
 
-				var colliders = Physics.OverlapSphere(rayOrigin.position, radius, -1, QueryTriggerInteraction.Collide);
-				foreach (var collider in colliders)
+				var totalColliders = Physics.OverlapSphereNonAlloc(rayOrigin.position, radius, s_CachedColliders, -1, QueryTriggerInteraction.Collide);
+				var colliderIndex = 0;
+
+				while (colliderIndex < totalColliders)
```
mtschoen-unity: Can be a for loop

AndrewTHEManeri:

Scripts/Core/EditorVR.UI.cs
```diff
@@ -129,7 +129,14 @@ void SetManipulatorsVisible(ISetManipulatorsVisible setter, bool visible)
 
 			bool GetManipulatorDragState()
 			{
-				return m_ManipulatorControllers.Any(controller => controller.manipulatorDragging);
+				foreach (var currentController in m_ManipulatorControllers)
+				{
+					if (currentController.manipulatorDragging)
+					{
```
dunity: Remove if  curly braces

Scripts/Core/EditorVR.UI.cs
```diff
@@ -129,7 +129,14 @@ void SetManipulatorsVisible(ISetManipulatorsVisible setter, bool visible)
...
+						return true;
+					}
+				}
+				return false;
```
dunity: Add line between curly brace & return

Scripts/Modules/DeviceInputModule.cs
```diff
@@ -67,26 +68,26 @@ void OnDestroy()
...
-			foreach (var inputControl in removeList)
+			foreach (var inputControl in s_RemoveList)
 			{
 				if (!inputControl.provider.active)
 					ResetControl(inputControl);
 
 				m_LockedControls.Remove(inputControl);
 			}
-
+			s_RemoveList.Clear();
```
dunity: Add line between curly brace clear call

Scripts/Utilities/ComponentUtils.cs
```diff
@@ -0,0 +1,27 @@
...
+		{
+			var foundComponent = default(T);
+			go.GetComponents<T>(s_RetrievalList);
+			if (s_RetrievalList.Count > 0)
+			{
+				foundComponent = s_RetrievalList[0];
+				s_RetrievalList.Clear();
+			}
+			return foundComponent;
```
dunity: Add line between curly brace & return

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -146,6 +149,7 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 						m_HoverGameObjects.Remove(directRayOrigin);
 					}
 				}
+				s_TempHovers.Clear();
```
dunity: Add line between curly brace & clear

mtschoen-unity:
Took the hatchet to this after checking with @AndrewTHEManeri about it. Biggest wins were:

- Removing lambda expression from ForEachProxyRay and cacheing all delegate parameters used in per-frame methods. Basically the deal here is that even if you're not using an anonymous closure, we box the method as a delegate when using it as a method argument. So the one use of IForEachProxyRay and calls to ProcessInput needed this cacheing
- I had an old GC pass that I did to the input prototype from ages ago that actually pools events and gets rid of some per-frame garbage generated in sending those events. I haven't tested on Vive, so there might be some stuff we can kill in ViveInputToEvents
- More GetComponent goodness
- The use of a Dictionary was hurting us in TransformTool. I'm not sure what the deal is, but sometimes TryGetValue and Contains allocate, but sometimes they don't.

I also found and fixed some interaction bugs (mostly null refs) that only show up when the profiler is killing performance.

The last thing I did was to eliminate the recursion in NestedListViewController and its descendants. This wasn't generating garbage, but was spamming the profiler with method calls, so profiling is now about 2x as fast with those nested views open. I can't really speak to whether the runtime performance is any different, because it's doing approximately the same thing (pushing and popping data to a stack when making a recursive call) but it's doing it on the heap now. In any case, it will make our lives easier as we try to squash perf bugs. As always, converting from a recursive to an iterative strategy takes a little more code, but thankfully the inner loop doesn't really have to change.

I had to change the merge target since I merged with latest dev.

stella3d:
is it not possible to use `Physics.RaycastNonAlloc` ?

mtschoen-unity:
@stella3d are you referring to PhysicsRaycaster.Raycast? It probably could/should use NonAlloc, but it's UI code, so we'd have to make users use a custom UI DLL

stella3d:
yeah, that's what i meant.  dang :(

stella3d:
code looks good to me - i think i know where the SmoothCamera allocations are coming from, if you're ok with me tackling that in this branch ?

mtschoen-unity:
Do it :+1:

Tools/TransformTool/TransformTool.cs
```diff
@@ -737,7 +732,24 @@ void UpdateManipulatorAction()
...
+
+        GrabData GrabDataForNode(Node node)
+        {
+            return node == Node.LeftHand ? m_LeftGrabData : m_RightGrabData;
+        }
+
+        GrabData GrabDataForRayOrigin(Transform rayOrigin)
+        {
+            var grabData = m_LeftGrabData;
+            if (grabData != null && grabData.rayOrigin == rayOrigin)
```
AndrewTHEManeri: Is this correct?  If the  leftGrabData rayOrigin is the rayOrigin, we return the rightGrabData?

mtschoen-unity: Good catch. This was a refactor gone bad.

P.S. @stella3d sorry if you had already committed on top of the current state. It should auto-merge just fine if you pull my changes, then you should be able to push.

AndrewTHEManeri:
I approve

AndrewTHEManeri:

PR #349, opened 2017-10-06
Vive proxy feedback and animation

mtschoen-unity:
### Purpose of this PR
Get Vive controllers up to speed with Controllers project

### Testing status

Tested on Vive and Rift using SteamVR

### Technical risk

Low: Mostly prefab changes and adding assets

### Comments to reviewers

The process was actually pretty painless once I had the Vive controller model with all of its proper pivots and stuff set up. I actually found that the values in the json file didn't exactly match what we needed, and there was some odd Y/Z confusion that I'm curious to track down. In any case, what we now have is a pair of prefabs for Vive and Touch that will map to the OpenVR tracking locations and has the proper animator/feedback setup.

One of the interesting things that had to change was to accommodate multiple tooltips per object, in the case of the trigger and grips, where there are different target destinations depending on which hand you're holding the Vive controller in. Come to think of it, the changes to TooltipModule and TwoHandedProxyBase to operate on multiple components aren't necessary, since we clear out the extra tooltips once the prefab has been assigned to a hand. However, I can still see a situation where one gameobject might have the capability of displaying two different tooltips at once, in which case we'd still want to support this situation. Before adding the step that deletes the excess toolitp components, the double-tooltip setup worked just fine.

You also might notice that I had to switch hands  in ViveInputToEvents, which might have explained why, in the past, we had backwards Touch controllers under OpenVR.

Also of interest is the postAnimate event, which is what allows me to do the custom logic for the Vive trackpad. The way the SteamVR examples do it is to only show the sphere while you're touching the trackpad, and to only rotate the trackpad if you're pressing on the button. Both of these cases where either the visibility of the feedback object changes or the feedback is contingent on more than one source of input (i.e. rotate but only if pad is pressed) aren't something I think the ButtonObject/ProxyAnimator system should handle generically, and they're easily done as overrides after "regular" animations are done.


Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -8,18 +9,34 @@
...
 		public Vector3 rotationOffset;
 
-		public void Reset()
+		public void ResetOffsets()
 		{
 			positionOffset = Vector3.zero;
 			rotationOffset = Vector3.zero;
 		}
+
+		public void ResetPositionOffset()
```
AndrewTHEManeri: Just make the position and rotation offsets properties and set them to zero directly?  These methods seem redundant

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -74,34 +73,56 @@ public virtual bool hidden
...
 		public virtual void Awake()
 		{
 			m_LeftHand = ObjectUtils.Instantiate(m_LeftHandProxyPrefab, transform).transform;
 			m_RightHand = ObjectUtils.Instantiate(m_RightHandProxyPrefab, transform).transform;
 			var leftProxyHelper = m_LeftHand.GetComponent<ProxyHelper>();
 			var rightProxyHelper = m_RightHand.GetComponent<ProxyHelper>();
 
-			m_ProxyMeshRoots.Add(leftProxyHelper.meshRoot);
-			m_ProxyMeshRoots.Add(rightProxyHelper.meshRoot);
+			foreach (var tooltip in leftProxyHelper.rightTooltips)
```
AndrewTHEManeri: Should probably encapsulate this so the we don't have to have nearly the same code twice in a row

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -37,14 +36,14 @@ abstract class TwoHandedProxyBase : MonoBehaviour, IProxy, IFeedbackReceiver, IS
...
-		readonly List<Material> m_Materials = new List<Material>();
 		readonly List<ProxyFeedbackRequest> m_FeedbackRequests = new List<ProxyFeedbackRequest>();
 
 		protected Dictionary<Node, Transform> m_RayOrigins;
 
-		List<Transform> m_ProxyMeshRoots = new List<Transform>();
+		bool m_Hidden;
 
-		readonly Dictionary<Node, Dictionary<VRInputDevice.VRControl, ProxyHelper.ButtonObject>> m_Buttons = new Dictionary<Node, Dictionary<VRInputDevice.VRControl, ProxyHelper.ButtonObject>>();
+		readonly Dictionary<Node, Dictionary<VRInputDevice.VRControl, List<ProxyHelper.ButtonObject>>> m_Buttons =
```
AndrewTHEManeri: Is this to handle button-combinations?

Scripts/Proxies/ProxyAnimator.cs
mtschoen-unity: Yeah, they're already public so looking back I'm not sure why I didn't do this originally

Scripts/Proxies/TwoHandedProxyBase.cs
mtschoen-unity: This specifically came up because the Vive grip button is two physical buttons, which need separate renders and transforms, but are still considered one "input control."

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -62,10 +62,13 @@ public class ButtonObject
 		Transform m_FieldGrabOrigin;
 
 		[SerializeField]
-		Transform m_MeshRoot;
+		ButtonObject[] m_Buttons;
 
 		[SerializeField]
-		ButtonObject[] m_Buttons;
+		Tooltip[] m_LeftTooltips;
```
dunity: Left and Right Tooltips should be removed from ProxyHelper, and into ViveProxy.

AndrewTHEManeri:

PR #352, opened 2017-10-13
Complete Feedback Pass

mtschoen-unity:
### Purpose of this PR
The first feedback PR was just to flesh out IRequestFeedback and FeedbackReceivers. This actually implements feedback across all EVR systems (at least all that I could think of). Here is the current list of classes that implements IFeedbackReceiver:

- MainMenu
- RadialMenu
- ToolsMenu
- MultipleRayInputModule
- AnnotationTool
- CreatePrimitiveTool
- LocomotionTool
- SelectionTool
- TransformTool
- VacuumTool
- WorkspaceUI
- MiniWorldWorkspace
- AssetGridItem

In practice, this is clearly too much feedback. There are one or two situations where tooltips can run into each other, and even when they don't, the controller kind of looks like Katamari when showing tooltips on all the buttons. We also discussed that the current position/orientations for the tooltips are not ideal. I didn't spend a lot of time on them knowing that we would want to see the full extent of feedback in the system before deciding how to arrange the tooltips for the controllers.

### Testing status

Tested in Adventure test scene and blank scene, with a variety of workspace positions and combinations.

### Technical risk

Low. No existing functionality is changed except to fix a bug in how TwoHandedProxyBase deals with feedback priority. Everything else is just added feedback calls and slight tweaks to input logic. 

### Comments to reviewers

I had to make the tooltips write depth to get them visible over the miniworld. I didn't want to mess with the materials too much, but the highlight outline currently doesn't draw over miniworld content, which looks a bit odd. The issue of transparent objects drawing over the miniworld should be addressed soon. I think the only way to do it is with multiple cameras/layers. One camera draws the scene, a number of Miniworld cameras draw in the next layer, and then our controllers, menus, tooltips, and workspaces draw in a third layer on top of everything. Currently, miniworlds draw after the scene camera, which is why transparent UI (like world scale visuals) draws underneath miniworld content.

I ended up adding AnnotationTool to the tools namespace, so you might want to view those changes with ?w=1


Menus/MainMenu/MainMenu.cs
```diff
@@ -363,6 +376,24 @@ public void RemoveSettingsMenuItem(ISettingsMenuItemProvider provider)
             }
             provider.settingsMenuItemInstance = null;
         }
+
+        void ShowFeedback()
+        {
+            var tooltipText = proxyType == typeof(ViveProxy) ? "Press to Rotate Menu" : "Rotate Menu";
```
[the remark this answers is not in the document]
mtschoen-unity: While this is something we've done specifically to accomodate the Vive, we also implement this "either or" behavior directly in MainMenu as well. If we are to generalize this capability, I'd also like to do it at the input level as well to remain consistent.

Menus/RadialMenu/RadialMenu.cs
```diff
@@ -2,18 +2,22 @@
...
 using System.Collections.Generic;
 using UnityEditor.Experimental.EditorVR.Core;
+using UnityEditor.Experimental.EditorVR.Proxies;
+using UnityEditor.Experimental.EditorVR.Utilities;
 using UnityEngine;
 using UnityEngine.InputNew;
 
 namespace UnityEditor.Experimental.EditorVR.Menus
 {
+    using BindingDictionary = Dictionary<string, List<VRInputDevice.VRControl>>;
```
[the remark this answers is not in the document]
mtschoen-unity: I was overthinking things. I forgot that you can just make an empty class that extends Dictionary as a template. I'll do that.

mtschoen-unity:
I've addressed the class alias and "stuck on" spatial scroll feedback. I'm not sure about making the tool-specific code an extension method or API because I wouldn't want to suggest that it was required. It's such simple code that I don't mind copy/pasting it everywhere.

As for making tooltip lines draw on top, it would involve having them write depth, which would make them "cut through" geometry. If you're just talking about drawing on top of controllers, we could turn off Z testing, but since tooltips *do* z-test, you might end up just seeing the dotted line in certain situations.

Menus/MainMenu/MainMenu.cs
```diff
@@ -363,6 +374,24 @@ public void RemoveSettingsMenuItem(ISettingsMenuItemProvider provider)
...
+            List<VRInputDevice.VRControl> controls;
+            if (m_Controls.TryGetValue("FlickFace", out controls))
```
jono-unity: The hardcoded strings here are freaking me out. Schoen, Dylan, and I were talking about adding a scriptable object field to action source on the action map, which would contain the control name, tooltip string, and any other data we need associated with the affordance.

mtschoen-unity: It sounds like we pretty much want to make everything about Feedback serializable. This is fine for our implementation, but I want to make sure this hard-coded path still works, since there is less friction to just getting some feedback up and running this way. Same goes with the repeated code: If we are just reading serialized tooltips from actions, that will all go away.

I am a little worried about coupling proxy feedback to actionmaps, but I think because this isn't all feedback ever, it makes sense that controller feedback relies on the input system.

I think we should have one last group conversation about the feedback-in-actionmaps plan, specifically to get input from @AndrewTHEManeri before proceeding.

mtschoen-unity: Tooltip flicker fixed. Good catch 👍

Menus/MainMenu/MainMenu.cs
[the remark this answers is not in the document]
AndrewTHEManeri: Can we just make this 'rotate menu' in all cases then?  I think that's explanatory enough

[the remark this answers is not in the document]
mtschoen-unity: I've seen this be an issue when introducing people to EVR, so that's why I went out of my way to say "press" on Vive. What do other people think.

Tools/AnnotationTool/AnnotationTool.cs
```diff
@@ -13,816 +11,842 @@
...
-        else
+
+        public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
         {
-            // Secondary hand uses brush size input to do undo/redo
-            var value = annotationInput.changeBrushSize.value;
-            if (proxyType == typeof(ViveProxy))
+            var annotationInput = (AnnotationInput)input;
+
+            var isHeld = annotationInput.draw.isHeld;
```
dunity: Cache off "annotationInput.draw" and use throughout function

Tools/AnnotationTool/AnnotationTool.cs
```diff
@@ -13,816 +11,842 @@
...
+        IEnumerator SetAnnotationPointerVisibility(bool visible)
         {
-            annotationPointerTransform.localScale = Vector3.Lerp(currentScale, targetScale, timeDiff / transitionTime);
-            timeDiff = Time.time - startTime;
-            yield return null;
-        }
+            if (!m_AnnotationPointer)
+                yield break;
+
+            const float transitionTime = 0.1875f;
```
dunity: rename to kTransitionTime

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -30,9 +31,10 @@ public class RaycastSource : ICustomActionMap
             public bool hasObject { get { return currentObject != null && (s_LayerMask & (1 << currentObject.layer)) != 0; } }
 
             public ActionMap actionMap { get { return m_Owner.m_UIActionMap; } }
-
             public bool ignoreLocking { get { return false; } }
 
+            readonly List<ProxyFeedbackRequest> m_ScrollFeedback = new List<ProxyFeedbackRequest>();
```
dunity: private readonly above publics

Workspaces/MiniWorldWorkspace/MiniWorldWorkspace.cs
```diff
@@ -431,6 +482,88 @@ protected override void OnDestroy()
...
+                        hideExisting = hideExisting
+                    };
+
+                    this.AddFeedbackRequest(request);
+                    requests.Add(request);
+                }
+            }
+        }
+
+        void ShowLeftMoveFeedback()
```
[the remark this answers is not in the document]
AndrewTHEManeri: Yeah, these one-line wrappers would work with a const string just as well, you've convinced me

Menus/MainMenu/MainMenu.cs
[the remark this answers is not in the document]
AndrewTHEManeri: How about "Press Edge To Rotate Menu" for both?

Menus/MainMenu/MainMenu.cs
AndrewTHEManeri: I agree with a more serialized approach to this - does it need to be this PR or can it be a subsequent one?

mtschoen-unity:
I'd say the serialized stuff should wait until we land Dylans PR which changes how animation definitions are serialized

Tools/AnnotationTool/AnnotationTool.cs
mtschoen-unity: We're not doing that for local consts

Menus/MainMenu/MainMenu.cs
mtschoen-unity: I'd say definitely not. It will overlap with Dylan's PR, so we should wait until that stabilizes.

Menus/MainMenu/MainMenu.cs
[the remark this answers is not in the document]
mtschoen-unity: I'm still of a mind to keep it as-is. Let's take this discussion offline to reach consensus.

mtschoen-unity:
@AndrewTHEManeri and I discussed the "Press to rotate" and one-liner method issues, and I've addressed them in 4d47fa7

AndrewTHEManeri:

PR #353, opened 2017-10-13
Hide MainMenu if it overlaps the tools menu

mtschoen-unity:
Timoni requested this yesterday as she was looking at controller visuals. While I was touching the local method use lists, I made the rest of them static, hence the seemingly unrelated rename changes.


AndrewTHEManeri:
Good, manual merge successful, performing github merge

PR #355, opened 2017-10-16
Automatic proxy (semi)transparency support

dunity:
### Purpose of this PR
Add automatic animated transparency support for proxies.  Static, non-interactive elements remain semi-transparent.  Affordances, proxy elements that can be interacted with (buttons, analog, etc) become opaque when feedback is being displayed on them (for a brief period of time), or when the user shakes their hands back and forth (which makes the body meshes opaque as well for a brief time).

### Testing status

Tested in Various scenes and image effect environments, appeared as expected.

### Technical risk

Low - The changes herein don't affect many areas outside the TwoHandedProxyBase. Care has been given, to avoid altering any of the ways in which the feedback system has been implemented in THPB.  The shake-to-reveal functionality also allows for feedback to be added/removed as it did previously; in my tests there were no conflicts with the current way in which this functions alongside feedback in THPB.

### Comments to reviewers

I separated, into another branch, the merging of the proxy animator data into ProxyAffordanceMaps (via a new AnimationDefinition data class).  That code now resides in improvements/dylanu/controller-transparency-animator-support; to be PR'd after this.  I broke out the code into that branch, In order to get this merged, and prevent more constant code-drift from this AffordanceMap architecture, due to additional time needed to finish the animator merge and support in AffordanceMaps.
Due to the nature of supporting (semi)transparency in the the shaders/materials for proxies, there is the usual "depth popping" visual artifact on some controller elements at certain angles.  An additional z-only prepass can be added to negate the effect if too distracting, coming at the expense of additional draws.  I plan on mocking this up for perf testing, and design approval later.

PR #356, opened 2017-10-19
Automatic proxy (semi)transparency support + ProxyAnimator support

dunity:
### Purpose of this PR
Add automatic animated transparency support for proxies.  Static, non-interactive elements remain semi-transparent.  Affordances, proxy elements that can be interacted with (buttons, analog, etc) become opaque when feedback is being displayed on them (for a brief period of time), or when the user shakes their hands back and forth (which makes the body meshes opaque as well for a brief time).

_This PR is based off of the original controller-transparency PR, and also includes ProxyAnimator arch refactors to accommodate the new ProxyAffordanceMap structure._ https://github.com/Unity-Technologies/EditorVR/pull/355

### Testing status

Tested in Various scenes and image effect environments, appeared as expected.

### Technical risk

Low - The changes herein don't affect many areas outside the TwoHandedProxyBase. Care has been given, to avoid altering any of the ways in which the feedback system has been implemented in THPB.  The shake-to-reveal functionality also allows for feedback to be added/removed as it did previously; in my tests there were no conflicts with the current way in which this functions alongside feedback in THPB.

### Comments to reviewers

I separated, into another branch, the merging of the proxy animator data into ProxyAffordanceMaps (via a new AnimationDefinition data class).  That code now resides in improvements/dylanu/controller-transparency-animator-support; to be PR'd after this.  I broke out the code into that branch, In order to get this merged, and prevent more constant code-drift from this AffordanceMap architecture, due to additional time needed to finish the animator merge and support in AffordanceMaps.
Due to the nature of supporting (semi)transparency in the the shaders/materials for proxies, there is the usual "depth popping" visual artifact on some controller elements at certain angles.  An additional z-only prepass can be added to negate the effect if too distracting, coming at the expense of additional draws.  I plan on mocking this up for perf testing, and design approval later.


Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+
+                                visibilityDefinition.materialsAndAssociatedColors = materialsAndAssociatedColors;
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (!affordancesDefinedInMap)
+                m_AffordanceMap.AffordanceDefinitions = generatedAffordanceDefinitions.ToArray();
```
AndrewTHEManeri: If there are affordances defined in the map, but not for every control, are those generated affordances being lost here?

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            {
+                    var animateFromColor = materialAndAssociatedColors.firstElement.GetColor(shaderColorPropety); // Get current color from material
+                    var animateToColor = isVisible ? materialAndAssociatedColors.secondElement : materialAndAssociatedColors.thirdElement; // (second)original or (third)hidden color(alpha/color.a)
+                    materialAndAssociatedColors.fourthElement = animateFromColor;
+                    materialAndAssociatedColors.fifthElement = animateToColor;
+            }
+
+            while (currentAmount < kTargetAmount)
+            {
+                var smoothedAmount = MathUtilsExt.SmoothInOutLerpFloat(currentAmount += Time.unscaledDeltaTime * speedScalar);
```
AndrewTHEManeri: It's not as fancy but let's put the code that actually lets this loop complete on its own line

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            {
+                var animateFromAlpha = materialAndAssociatedColors.firstElement.GetFloat(shaderAlphaPropety); // Get current alpha from material
+                var animateToAlpha = isVisible ? materialAndAssociatedColors.secondElement.a : materialAndAssociatedColors.thirdElement.a; // (second)original or (third)hidden color(alpha/color.a)
+                materialAndAssociatedColors.fourthElement = Color.white * animateFromAlpha; // Encode the alpha for the FROM color value, color.a
+                materialAndAssociatedColors.fifthElement = Color.white * animateToAlpha; // // Encode the alpha for the TO color value, color.a
+            }
+
+            while (currentAmount < kTargetAmount)
+            {
+                var smoothedAmount = MathUtilsExt.SmoothInOutLerpFloat(currentAmount += Time.unscaledDeltaTime * speedScalar);
```
AndrewTHEManeri: same thing here

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            }
+
+            const float kTargetAmount = 1f;
+            const float kHiddenValue = 0.25f;
+            var speedScalar = isVisible ? k_FadeInSpeedScalar : k_FadeOutSpeedScalar;
+            var currentAmount = 0f;
+            var shaderColorPropety = bodyVisibilityDefinition.colorProperty;
+            while (currentAmount < kTargetAmount)
+            {
+                var smoothedAmount = MathUtilsExt.SmoothInOutLerpFloat(currentAmount += Time.unscaledDeltaTime * speedScalar);
```
AndrewTHEManeri: and here

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            foreach (var kvp in m_BodyMaterialOriginalColorMap)
+            {
+                var originalColor = kvp.Value.originalValue;
+                kvp.Key.SetColor(shaderColorPropety, isVisible ? originalColor : new Color(originalColor.r, originalColor.g, originalColor.b, kHiddenValue));
+            }
+        }
+
+        IEnumerator AnimateBodyAlphaVisibility(bool isVisible)
+        {
+            var bodyVisibilityDefinition = m_AffordanceMap.bodyVisibilityDefinition;
```
AndrewTHEManeri: A lot of these loops are extremely visible and even rely on the same kind of templated data -  any way we can combine these functions?

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -27,26 +30,36 @@ public void Apply(Transform transform)
     [SerializeField]
     ActionMap m_ProxyActionMap;
 
-    ProxyHelper.ButtonObject[] m_Buttons;
+    [SerializeField]
+    ProxyAffordanceMap m_AffordanceMap;
```
AndrewTHEManeri: Should this just come from the proxy helper rather than being another serialized reference?  We have to maintain this link in here and the proxy UI right now

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -94,16 +107,24 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 
         for (var i = 0; i < length; i++)
         {
-            var button = m_Buttons[i];
+            var affordance = m_Affordances[i];
             var control = m_Controls[i];
-            var info = m_TransformInfos[button.transform];
+            var affordanceDefinition = m_AffordanceDefinitions.Where(x => x.control == affordance.control).FirstOrDefault();
```
AndrewTHEManeri: don't use linq here, this is per-frame

Tools/TransformTool/TransformTool.cs
```diff
@@ -787,8 +787,8 @@ void ShowFeedback(List<ProxyFeedbackRequest> requests, string controlName, strin
                         control = id,
                         tooltipText = tooltipText,
                         priority = 1,
-                        hideExisting = hideExisting
-                    };
+	                    suppressExisting = suppressExisting
```
AndrewTHEManeri: tabs

Tools/AnnotationTool/AnnotationTool.cs
```diff
@@ -266,7 +266,7 @@ void Start()
                             control = id,
                             priority = 1,
                             tooltipText = tooltipText,
-                            hideExisting = hideExisting
+	                        suppressExisting = suppressExisting
```
AndrewTHEManeri: tab

Scripts/Proxies/Vive/ViveProxy.cs
```diff
@@ -58,12 +60,12 @@ public override void Awake()
...
 
-        static void PostAnimate(ProxyHelper.ButtonObject[] buttons, Dictionary<Transform, ProxyAnimator.TransformInfo> transformInfos, ActionMapInput input)
+        static void PostAnimate(Affordance[] affordances, ProxyAffordanceMap.AffordanceDefinition[] affordanceDefinitions, Dictionary<Transform, ProxyAnimator.TransformInfo> transformInfos, ActionMapInput input)
         {
             var proxyInput = (ProxyAnimatorInput)input;
-
-            foreach (var button in buttons)
+            foreach (var button in affordances)
             {
+                var affordanceDefinition = affordanceDefinitions.Where(x => x.control == button.control).FirstOrDefault().animationDefinition;
```
AndrewTHEManeri: this also happens through ProcessInput (and thus per-frame), so let's avoid linq here

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -38,35 +43,36 @@ abstract class TwoHandedProxyBase : MonoBehaviour, IProxy, IFeedbackReceiver, IS
...
+        ProxyFeedbackRequest m_SemitransparentLockRequest;
+        float m_ShakeFrequency;
+        Vector3 m_PreviousLeftHandPosition;
+        Vector3 m_PreviousRightHandPosition;
 
-        public Transform leftHand
-        {
-            get { return m_LeftHand; }
-        }
+        readonly Dictionary<Node, Dictionary<VRInputDevice.VRControl, List<Affordance>>> m_Affordances =
```
AndrewTHEManeri: Anyway we can shrink this down to an affordanceDictionary or some such?

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +171,50 @@ public virtual IEnumerator Start()
...
         }
 
-        public virtual void OnDestroy() { }
-
         public virtual void Update()
         {
             if (active)
             {
-                m_LeftHand.localPosition = trackedObjectInput.leftPosition.vector3;
+                var movementDelta = -Time.unscaledDeltaTime * 4;
```
AndrewTHEManeri: magical number that seems like it does nothing?

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +171,50 @@ public virtual IEnumerator Start()
...
-                m_RightHand.localPosition = trackedObjectInput.rightPosition.vector3;
+                var rightLocalPosition = trackedObjectInput.rightPosition.vector3;
+                m_RightHand.localPosition = rightLocalPosition;
                 m_RightHand.localRotation = trackedObjectInput.rightRotation.quaternion;
+
+                if (m_SemitransparentLockRequest == null)
+                {
+                    movementDelta = Vector3.SqrMagnitude(leftLocalPosition - m_PreviousLeftHandPosition);
+                    movementDelta += Vector3.SqrMagnitude(rightLocalPosition - m_PreviousRightHandPosition);
+                    movementDelta *= Time.unscaledDeltaTime * 20;
```
AndrewTHEManeri: additional magic

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +171,50 @@ public virtual IEnumerator Start()
...
+                    if (movementDelta > 0.001f)
+                    {
+                        m_ShakeFrequency += movementDelta;
+                        if (m_ShakeFrequency > 0.1f)
+                        {
+                            var shakeRequest = new ProxyFeedbackRequest
+                            {
+                                control = VRInputDevice.VRControl.LocalRotation,
```
AndrewTHEManeri: Why rotation and not position?

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -277,14 +334,72 @@ void RemoveFeedbackRequest(ProxyFeedbackRequest request)
...
+            if (shakenVisibility)
+            {
+                // Left & right device affordances should be visible when the input-device is shaken
+                rightProxyRequestsExist = true;
+                leftProxyRequestsExist = true;
+            }
+            else if (m_FeedbackRequests.Count > 0)
+            {
+                // Find any visible feedback requests for each hand
+                rightProxyRequestsExist = m_FeedbackRequests.Any(x => x.Key.node == Node.RightHand && x.Key.visible);
```
AndrewTHEManeri: Seems like this can happen often enough we want to avoid the linq

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -27,26 +30,36 @@ public void Apply(Transform transform)
...
+    public Node node { get { return m_Node; } set { m_Node = value; m_RightHandedProxy = m_Node == Node.RightHand; } }
 
-    void Start()
+    internal event Action<Affordance[], AffordanceDefinition[], Dictionary<Transform, TransformInfo>, ActionMapInput> postAnimate;
+
+    public void Setup(Affordance[] affordances)
     {
-        m_Buttons = GetComponent<ProxyHelper>().buttons;
+        m_Affordances = affordances;
+        m_AffordanceDefinitions = m_AffordanceMap.AffordanceDefinitions;
```
mtschoen-unity: Can this second part happen in Awake or Start? It may seem excessive to add another method, but I don't want to give the impression that this somehow relies on m_Affordances being set.

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -27,26 +30,36 @@ public void Apply(Transform transform)
     [SerializeField]
     ActionMap m_ProxyActionMap;
 
-    ProxyHelper.ButtonObject[] m_Buttons;
+    [SerializeField]
+    ProxyAffordanceMap m_AffordanceMap;
+
+    Affordance[] m_Affordances;
+    ProxyAffordanceMap.AffordanceDefinition[] m_AffordanceDefinitions;
```
mtschoen-unity: ProxyAffordanceMap is redundant with your alias above. Also see my comment about promoting this to a first-level class.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -102,9 +71,40 @@ public class ButtonObject
...
+            origins.Add(previewOrigin);
+            origins.Add(fieldGrabOrigin);
+
+            if (m_ProxyUI)
+                m_ProxyUI.Setup(m_Affordances, origins);
+
+            if (m_ProxyAnimator)
+            {
+                m_ProxyAnimator.Setup(m_Affordances);
+                this.ConnectInterfaces(m_ProxyAnimator, this.rayOrigin);
```
mtschoen-unity: Is the order important here? I'd generally opt for setting interfaces first in case Setup needs injected functionality.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -102,9 +71,40 @@ public class ButtonObject
...
+
+        /// <summary>
+        /// Set the visibility of the renderers not associated with controls/input
+        /// </summary>
+        public bool bodyRenderersVisible { set { if (m_ProxyUI != null) m_ProxyUI.bodyVisible = value; } }
+
+        void Start()
+        {
+            // Setup ProxyUI
+            List<Transform> origins = new List<Transform>();
```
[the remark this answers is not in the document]
mtschoen-unity: I'm usually a fan of initializers for this kind of thing. I would actually recommend inlining it, and using brackets in the if statement below. That way, you only create the temp list if you end up using it.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -1,54 +1,17 @@
 ﻿#if UNITY_EDITOR
-using System;
-using UnityEditor.Experimental.EditorVR.UI;
+using System.Collections.Generic;
+using UnityEditor.Experimental.EditorVR.Utilities;
```
mtschoen-unity: Unused

Scripts/Helpers/ProxyHelper.cs
mtschoen-unity: Also "this." is redundant

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            var defaultAffordanceVisibilityDefinition = m_AffordanceMap.defaultAffordanceVisibilityDefinition;
+            var defaultAffordanceAnimationDefinition = m_AffordanceMap.defaultAnimationDefinition;
+            foreach (var proxyAffordance in affordances)
+            {
+                var renderers = proxyAffordance.renderer.GetComponentsInChildren<Renderer>(true);
+                if (renderers != null)
+                {
+                    // Setup animated color or alpha transparency for all materials associated with all renderers associated with the control
+                    ProxyAffordanceMap.AffordanceVisibilityDefinition visibilityDefinition;
+                    ProxyAffordanceMap.AffordanceAnimationDefinition animationDefinition;
```
mtschoen-unity: Can be declared in inner scope

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+
+            // Set affordances AFTER setting origins, as the origins are referenced when setting up affordances
+            m_ProxyOrigins = proxyOrigins;
+
+            // Clone the affordance map, in order to allow a single map to drive the visuals of many duplicate
+            // This also allows coroutine sets in the ProxyAffordanceMap to be performed simultaneously for n-number of devices in a proxy
+            m_AffordanceMap = Instantiate(m_AffordanceMap);
+
+            m_Affordances = affordances;
+            m_AffordanceRenderers = new List<Renderer>();
```
mtschoen-unity: can be readonly and initialized at declaration

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+                            var originalColor = materialClone.GetColor(bodyVisibilityDefinition.colorProperty);
+                            if (materialClone.HasProperty(k_ZWritePropertyName))
+                                materialClone.SetFloat(k_ZWritePropertyName, 1);
+
+                            m_BodyMaterialOriginalColorMap[materialClone] = new affordancePropertyTuple<Color>(originalColor, originalColor);
+                        }
+                    }
+                    break;
+                case VisibilityControlType.alphaProperty:
+                    string shaderAlphaPropety = bodyVisibilityDefinition.alphaProperty;
```
mtschoen-unity: var

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+        readonly Dictionary<Material, affordancePropertyTuple<Color>> m_BodyMaterialOriginalColorMap = new Dictionary<Material, affordancePropertyTuple<Color>>();
+        readonly Dictionary<Material, affordancePropertyTuple<float>> m_BodyMaterialOriginalAlphaMap = new Dictionary<Material, affordancePropertyTuple<float>>();
+
+        /// <summary>
+        /// Model containing original value, and values to "animate from", unique to each body MeshRenderer material.
+        /// Visibility of all objects in the proxy body are driven by a single AffordanceVisibilityDefinition,
+        /// as opposed to individual interactable affordances, which each have their own AffordanceVisibilityDefinition, which contains their unique value data.
+        /// This is a lightweight class to store that data, alleviating the need to duplicate an affordance definition for each body renderer as well.
+        /// </summary>
+        class affordancePropertyTuple<T>
```
mtschoen-unity: Use Tuple, or capitalize class name

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            var materialsAndColors = visibilityDefinition.materialsAndAssociatedColors;
+            var shaderColorPropety = visibilityDefinition.colorProperty;
+
+            if (materialsAndColors == null)
+                yield break;
+
+            // Setup animateFromColors using the current color values of each material associated with all renderers drawing this affordance
+            foreach (var materialAndAssociatedColors in materialsAndColors)
+            {
+                    var animateFromColor = materialAndAssociatedColors.firstElement.GetColor(shaderColorPropety); // Get current color from material
```
mtschoen-unity: Excessive indentation in this for loop

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+
+        /// <summary>
+        /// Setup this ProxyUI
+        /// </summary>
+        /// <param name="affordances">The ProxyHelper affordances that drive visual changes in the ProxyUI</param>
+        /// <param name="proxyOrigins">ProxyOrigins whose child renderers will not be controlled by the PRoxyUI</param>
+        public void Setup(Affordance[] affordances, List<Transform> proxyOrigins)
+        {
+            // Prevent multiple setups
+            if (m_ProxyUISetup)
```
mtschoen-unity: Comments on these if statements are redundant with error messages.

Also, I wonder if this is something we can/should do with an Editor utility (custom Editor or EditorWindow?), rather than a runtime validation. Thoughts, @AndrewTHEManeri?

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+
+            m_Affordances = affordances;
+            m_AffordanceRenderers = new List<Renderer>();
+
+            // If no custom affordance definitions are defined in the affordance map, they will be populated by new generated definitions below
+            var affordanceMapDefinitions = m_AffordanceMap.AffordanceDefinitions;
+            var affordancesDefinedInMap = affordanceMapDefinitions != null && affordanceMapDefinitions.Length > 0 && affordanceMapDefinitions[0] != null;
+
+            // If affordanceMapDefinitions is null, set the list below into the map after setup
+            List<AffordanceDefinition> generatedAffordanceDefinitions = new List<AffordanceDefinition>();
```
mtschoen-unity: var

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+                        };
+
+                        animationDefinition = new ProxyAffordanceMap.AffordanceAnimationDefinition
+                        {
+                            translateAxes = defaultAffordanceAnimationDefinition.translateAxes,
+                            rotateAxes = defaultAffordanceAnimationDefinition.rotateAxes,
+                            min = defaultAffordanceAnimationDefinition.min,
+                            max = defaultAffordanceAnimationDefinition.max,
+                            reverseForRightHand = defaultAffordanceAnimationDefinition.reverseForRightHand
+                        };
```
mtschoen-unity: I recommend making AffordanceVisibilityDefinition and AffordanceAnimationDefinition to avoid having to deep copy like this. Generally, I think we'll always want to deep copy these things around.

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+            }
+
+            affordancesVisible = false;
+            bodyVisible = false;
+
+            // Allow setting of affordance & body visibility after affordance+body setup is performed in the "affordances" property
+            m_ProxyUISetup = true;
+        }
+
+        IEnumerator AnimateAffordanceColorVisibility(bool isVisible, AffordanceDefinition definition)
```
mtschoen-unity: static

Scripts/Helpers/ProxyUI.cs
AndrewTHEManeri: The problem we'll run into there is that this function is instantiating the affordance map and making extra entries for affordances that are not customized.  Caching this data (but changing it when the source changes) would still require a runtime check

Scripts/Helpers/ProxyAffordanceMap.cs
```diff
@@ -0,0 +1,159 @@
...
+        public class AffordanceVisibilityDefinition
+        {
+            [SerializeField]
+            VisibilityControlType m_VisibilityType;
+
+            [SerializeField] // colorProperty field
+            string m_ColorProperty = "_Color"; // Consider custom inspector that only displays this if this visibility type is chosen
+
+            [SerializeField] // alphaProperty field
+            string m_AlphaProperty = "_Alpha"; // Consider custom inspector that only displays this if this visibility type is chosen
```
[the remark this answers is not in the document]
dunity: There isn't really a highlight state currently, just a normal & hidden state.  An affordance is only either visible or hidden currently.  I saw support for highlights as something we'd consider during the iconography pass, which could/would include additional arch to support displaying sprites "on" affordances/buttons... which will definitely require a group planning session before proceeding.  Due to potential shader variants that might/might-not support color changes for highlight, I feel that we should consider alternatives decoupled from shader driven approaches for the iconography pass.

Scripts/Helpers/ProxyAffordanceMap.cs
```diff
@@ -0,0 +1,159 @@
...
+            /// <summary>
+            /// The renderer & cloned materials associated with this affordance
+            /// Cloned materials associated with the renderer will have their properties animated by the ProxyUI
+            /// Element 1: Original material
+            /// Element 2: Original color, Color.a is used for alpha-only animation
+            /// Element 3: Hidden color, Color.a is used for alpha-only animation
+            /// Element 4: Animate FROM color (used in animated coroutines), color.a is used for alpha-only animation
+            /// Element 5: Animate TO color (used in animated coroutines), color.a is used for alpha-only animation
+            /// </summary>
+            public List<Tuple<Material, Color, Color, Color, Color>> materialsAndAssociatedColors { get; set; }
```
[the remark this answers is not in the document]
dunity: I went with VisualStateData... changes underway now

Scripts/Helpers/ProxyAffordanceMap.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        [Serializable]
+        public class AffordanceDefinition
+        {
+            [SerializeField]
+            VRInputDevice.VRControl m_Control;
+
+            [SerializeField]
+            AffordanceAnimationDefinition m_AnimationDefinition;
```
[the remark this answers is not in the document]
dunity: There are fallbacks for an entire definition to the default definition.  I feel that extending the system to support fallbacks of individual definition properties might be best suited for another task/PR.

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+using UnityEditor.Experimental.EditorVR.Utilities;
+using UnityEngine;
+using VisibilityControlType = UnityEditor.Experimental.EditorVR.Core.ProxyAffordanceMap.VisibilityControlType;
+using AffordanceDefinition = UnityEditor.Experimental.EditorVR.Core.ProxyAffordanceMap.AffordanceDefinition;
+
+namespace UnityEditor.Experimental.EditorVR.Proxies
+{
+    public class ProxyUI : MonoBehaviour
+    {
+        const float k_FadeInSpeedScalar = 4f;
```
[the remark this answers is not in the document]
dunity: Good call

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+
+            // Don't allow setup if affordances are invalid
+            if (affordances == null || affordances.Length == 0)
+            {
+                Debug.LogError("Affordances invalid when attempting to setup ProxyUI on : " + gameObject.name);
+                return;
+            }
+
+            // Prevent further setup if affordance map isn't assigned
+            if (m_AffordanceMap == null)
```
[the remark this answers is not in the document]
dunity: I kept these separate due to some discussions that gave rise to situations that could call for separate  maps for the ProxyUI and the ProxyAnimator.  I originally had the map populated via the ProxyHelper.

Scripts/Proxies/ProxyAnimator.cs
dunity: Converting to m_AffordanceMapOverride.  Will default to using the override on ProxyAnimator & ProxyUI, if assigned, or use the default affordance map assigned to the proxy helper.

Scripts/Helpers/ProxyAffordanceMap.cs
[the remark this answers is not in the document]
AndrewTHEManeri: I'm down with that

Scripts/Helpers/ProxyAffordanceMap.cs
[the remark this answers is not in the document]
AndrewTHEManeri: I think the visual definition of what something looks like should be captured in one place - not if that state is active, just what it looks like.  Not likely a deal-breaker for this PR - but if we imagine our users are going to add some pretty novel proxies and ways of visualizing them, it might also require pretty different methods of visualizing a highlight than our current methods.  I could certainly see that being the case for a rigged semi-transparent hand you see with something like the leapmotion.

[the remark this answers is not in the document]
mtschoen-unity: There is an opportunity for this in the implementation of the feedback receiver. If we want it to respond to feedback with different visuals, that is totally possible

Scripts/Proxies/Vive/ViveProxy.cs
```diff
@@ -58,12 +60,12 @@ public override void Awake()
 #endif
         }
 
-        static void PostAnimate(ProxyHelper.ButtonObject[] buttons, Dictionary<Transform, ProxyAnimator.TransformInfo> transformInfos, ActionMapInput input)
+        static void PostAnimate(Affordance[] affordances, ProxyAffordanceMap.AffordanceDefinition[] affordanceDefinitions, Dictionary<Transform, ProxyAnimator.TransformInfo> transformInfos, ActionMapInput input)
```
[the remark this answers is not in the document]
dunity: Good call

Scripts/Proxies/TwoHandedProxyBase.cs
dunity: When you say shrink it down, do you mean create a class that derives from Dictionary, etc?

dunity: Thoughts @mtschoen ?

Scripts/Helpers/ProxyUI.cs
dunity: As per our chat, you meant Struct here, right? @mtschoen-unity

dunity: They've been refactored into classes re a previous PR comment, but I remember our Struct chat.

Scripts/Proxies/TwoHandedProxyBase.cs
mtschoen-unity: Yeah, this is what I did for ButtonDictionary. If it's just local, you can also use a class alias which doesn't affect the assembly, but either way, some kind of alias is best when the type is this long.

Also, FYI, my Unity account is @mtschoen-unity

Scripts/Helpers/ProxyUI.cs
mtschoen-unity: yeah. I was thinking those two types would be structs. That way you don't need to worry about deep copies or shared references

dunity: If we go this route, we lose the following default field initializers in AffordanceVisibilityDefinition : m_ColorProperty, m_AlphaProperty, and m_HiddenColor.   We also lose the m_Max initialized value in AffordanceAnimationDefinition.  If these are converted to structs, are you suggesting that those values, and all others required for initialization, be moved to a constructor?  Those defaults are nice to have in the inspector when creating a new definition, especially for new users using the standard shader imo.  Thoughts @AndrewTHEManeri  & @mtschoen-unity

mtschoen-unity: Yeah, we should be able to handle the default values with the constructor. It's a bit of an annoying rule, but I'm sure there are reasons. I wouldn't say we should get rid of the defaults.

Though, I'm not sure if the serialized fields get those defaults when they're set in the constructor. If that doesn't work, then let's just leave them as classes.

AndrewTHEManeri: How about just making a copy/clone constructor?  If we use these as arguments (I know we do for the visibilitydefinition) they're heavier to copy as arguments than I would want

dunity:
@AndrewTHEManeri  re your general playtesting notes:
1: Triggering individual affordance visibility based on input/affordance interaction will be coming in the second round of changes.
2: Coupling affordance visibility timing to Tooltips is coming later in the first round changes as well.
3: This'll be addressed in round two changes as well (or sooner).
4: Shaking detection was refactored in order to more reliably trigger the full opacity state of the proxy(s).

AndrewTHEManeri:
Fair enough, one last request - Can we get some comments on AffordanceDefinition,AffordanceAnimationDefinition and AffordanceVisibilityDefinition.

I don't need everything documented (though I won't stop you if you want to do that), just would like people to be able to see a class description on what these do/what step of the feedback process they fill in.

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +174,51 @@ public virtual IEnumerator Start()
...
 
-                m_RightHand.localPosition = trackedObjectInput.rightPosition.vector3;
+                var rightLocalPosition = trackedObjectInput.rightPosition.vector3;
+                m_RightHand.localPosition = rightLocalPosition;
                 m_RightHand.localRotation = trackedObjectInput.rightRotation.quaternion;
+
+                if (m_SemitransparentLockRequest == null)
+                {
+                    const float minShakeMovementThreshold = 0.001f;
+                    const int movementDetectionIncreaseScalar = 125;
```
jono-unity: These values feel a little magical; maybe they should move to the top of the class, if not a settings object somewhere, particularly because currently the shake seems a little inconsistent / hard to trigger, and we might need to adjust these some more to find the best spot.

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +174,51 @@ public virtual IEnumerator Start()
...
+                    var movementDelta = Vector3.SqrMagnitude(leftLocalPosition - m_PreviousLeftHandPosition);
+                    movementDelta += Vector3.SqrMagnitude(rightLocalPosition - m_PreviousRightHandPosition);
+                    movementDelta *= Time.unscaledDeltaTime * movementDetectionIncreaseScalar;
+                    if (movementDelta > minShakeMovementThreshold)
+                    {
+                        m_ShakeFrequency += movementDelta;
+                        if (m_ShakeFrequency > 0.1f)
```
jono-unity: More magic.

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -184,7 +227,19 @@ public void AddFeedbackRequest(FeedbackRequest request)
             var proxyRequest = request as ProxyFeedbackRequest;
             if (proxyRequest != null)
             {
-                m_FeedbackRequests.Add(proxyRequest);
+                var hasKey = m_FeedbackRequests.ContainsKey(proxyRequest);
+                if (hasKey) // Update existing request/coroutine pair
```
jono-unity: Only use of `hasKey`; just inline it.

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -277,14 +338,84 @@ void RemoveFeedbackRequest(ProxyFeedbackRequest request)
...
+                m_SemitransparentLockRequest = request;
+            }
+
+            request.visible = true;
+
+            const float kShakenVisibilityDuration = 5f;
+            const float kShorterOpaqueDurationScalar = 0.125f;
+            float duration = request.proxyShaken ? kShakenVisibilityDuration : k_DefaultFeedbackDuration * kShorterOpaqueDurationScalar;
+            var currentDuration = 0f;
+            while (request != null && currentDuration < duration)
```
jono-unity: `request` is always non-null, and you're accessing its properties earlier here anyway. Could just start this coroutine with `if (request == null) yield break;`, but that shouldn't be necessary.

Scripts/Proxies/TwoHandedProxyBase.cs
AndrewTHEManeri: I think what the shake detection is ultimately missing here is that it is a two-part problem:
You want to detect great average velocity over a period of time while the average position over that period of time has not changed much.  The velocity tracker from the toolkit is very well suited for this kind of thing - @dunity would you mind me taking a crack at this?

Scripts/Modules/TooltipModule/Tooltip.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace UnityEditor.Experimental.EditorVR.UI
 {
-    sealed class Tooltip : MonoBehaviour, ITooltip, ITooltipPlacement
+    public sealed class Tooltip : MonoBehaviour, ITooltip, ITooltipPlacement
```
mtschoen-unity: hmmm. Do we want all this to be public? If I go back to Tooltip being internal, I need to also make Affordance, ProxyAnimator, and ProxyUI internal.  So far, we've been treating all public classes as API or something which will be used by third-party packages.

I think all of this stuff should be internal.

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -27,26 +28,53 @@ public void Apply(Transform transform)
...
+    public Node node
+    {
+        get
+        {
+            return m_Node;
+        }
+
+        set
+        {
+            m_Node = value; m_RightHandedProxy = m_Node == Node.RightHand;
```
mtschoen-unity: No need for m_Node. Just set m_RightHandedProxy in setter, remove getter.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -102,9 +71,40 @@ public class ButtonObject
...
+            origins.Add(previewOrigin);
+            origins.Add(fieldGrabOrigin);
+
+            if (m_ProxyUI)
+                m_ProxyUI.Setup(m_AffordanceMap, m_Affordances, origins);
+
+            if (m_ProxyAnimator)
+            {
+                m_ProxyAnimator.Setup(m_AffordanceMap, m_Affordances);
+                this.ConnectInterfaces(m_ProxyAnimator, this.rayOrigin);
```
mtschoen-unity: remove this. before rayOrigin

Scripts/Proxies/ProxyAnimator.cs
```diff
@@ -27,26 +28,53 @@ public void Apply(Transform transform)
...
+            m_AffordanceMapOverride = affordanceMap;
+
+        m_Affordances = affordances;
+        m_AffordanceDefinitions = m_AffordanceMapOverride.AffordanceDefinitions;
     }
 
     public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeControl)
     {
-        if (m_Buttons == null)
+        if (m_Affordances == null)
```
mtschoen-unity: Probably not necessary after we combine ProxyUI and ProxyHelper

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -159,23 +174,51 @@ public virtual IEnumerator Start()
...
+                        if (m_ShakeFrequency > 0.1f)
+                        {
+                            var shakeRequest = new ProxyFeedbackRequest
+                            {
+                                control = VRInputDevice.VRControl.LocalPosition,
+                                node = Node.None,
+                                tooltipText = null,
+                                suppressExisting = true,
+                                proxyShaken = true
+                            };
```
mtschoen-unity: Can we re-use this request? Construct once and add over and over again? Come to think of it, I think we can do this with most feedback. Or maybe we just want to formalize into a FeedbackPool?

Scripts/Proxies/TwoHandedProxyBase.cs
```diff
@@ -277,14 +339,83 @@ void RemoveFeedbackRequest(ProxyFeedbackRequest request)
...
+            }
+
+            rightAffordanceRenderersVisible = rightProxyRequestsExist;
+            leftAffordanceRenderersVisible = leftProxyRequestsExist;
+
+            rightBodyRenderersVisible = shakenVisibility;
+            leftBodyRenderersVisible = shakenVisibility;
+        }
+
+        IEnumerator MonitorFeedbackRequestLifespan(ProxyFeedbackRequest request)
```
mtschoen-unity: On second review, I'm not sure I like this coupled to all feedback definitions. We may find that there are more situations where we want to monitor feedback, but for now I would rather this was just a single coroutine that we start when detect a shake that adds the feedback (if not already added), waits a period of time, and then removes it. This avoids turning the feedback list into a dictionary.

Scripts/Proxies/Vive/ViveProxy.cs
```diff
@@ -58,12 +59,21 @@ public override void Awake()
...
             {
+                AffordanceAnimationDefinition affordanceAnimationDefinition = null;
+                foreach (var definition in affordanceDefinitions)
+                {
+                    if (definition.control == button.control)
+                    {
+                        affordanceAnimationDefinition = definition.animationDefinition;
+                        break;
+                    }
+                }
```
mtschoen-unity: I'd like to refactor this into a dictionary, or a list of Tuples to avoid having to do this lookup here. Does that make sense, or will you have to do this n^2 task beforehand to match up the control IDs?

This makes me think that we should do this mapping at Start() and always deal with a Dictionary of control-to-definition or list of definitions.

Scripts/Helpers/ProxyAffordanceMap.cs
```diff
@@ -0,0 +1,37 @@
+﻿#if UNITY_EDITOR
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR.Core
+{
+    [CreateAssetMenu(menuName = "EditorVR/EXR Proxy Affordance Map", fileName = "NewProxyAffordanceMap.asset")]
```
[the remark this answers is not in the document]
dunity: good call

Scripts/Proxies/TwoHandedProxyBase.cs
dunity: As per our slack discussion, this covers all feedback countdown durations, not just shake.  So leaving in-place for now, as this is the means of ticking display duration for each request (checked against in UpdateVisibility for each hand).

Scripts/Proxies/Vive/ViveProxy.cs
dunity: Just pushed list of Tuple <ProxyFeedbackRequest, Coroutine> refactor to THPB.  Added an alias too.

Scripts/Helpers/ProxyHelper.cs
```diff
@@ -102,9 +71,40 @@ public class ButtonObject
...
+        /// Set the visibility of the renderers associated with affordances(controls/input)
         /// </summary>
-        public ButtonObject[] buttons { get { return m_Buttons; } }
+        /// Null checking before setting, as upon EXR setup, in Awake(), m_ProxyUI is null, even though it has been assigned in the inspector
+        public bool affordanceRenderersVisible { set { if (m_ProxyUI != null) m_ProxyUI.affordancesVisible = value; } }
+
+        /// <summary>
+        /// Set the visibility of the renderers not associated with controls/input
+        /// </summary>
+        public bool bodyRenderersVisible { set { if (m_ProxyUI != null) m_ProxyUI.bodyVisible = value; } }
```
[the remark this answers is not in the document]
dunity: Agreed on the merge call

Scripts/Helpers/ProxyUI.cs
```diff
@@ -0,0 +1,549 @@
...
+        {
+            public T originalValue { get; private set; }
+            public T animateFromValue { get; set; }
+
+            public AffordancePropertyTuple(T originalValue, T animateFromValue)
+            {
+                this.originalValue = originalValue;
+                this.animateFromValue = animateFromValue;
+            }
+        }
```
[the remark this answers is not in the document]
dunity: IMO this better conveys the animateFromValue than the generic tuple would.  I'm good with going with either approach though.

Scripts/Modules/TooltipModule/Tooltip.cs
dunity: @AndrewTHEManeri  thoughts?  I'm leaning toward internal as well.

AndrewTHEManeri: Internal

Scripts/Proxies/ProxyAnimator.cs
dunity: I'll leave it for now, till that merge happens (likely after this PR).

AndrewTHEManeri:
No further feedback here - I'd rather wait until some of the bugfixes like the controller ghost are merged in before I approve this

dunity:
I've addressed the outstanding issues/comments that I'm aware of relating to this PR.  Lemme know if there are any I might have missed, or if additional work is needed before a dev merge can occur. @AndrewTHEManeri  @mtschoen-unity

mtschoen-unity:
I ported over two of the simple fixes I had done in my controller-orientation branch
- Fix an InvalidOperationException due to me not cleaning up the old ClearFeedbackRequests implementation
- Fix tooltip queue re-using every tooltip object instance

I've fixed the left-hand-animates-both-hands bug in controller-orientation. It's a result of bringing feedback execution into ProxyUI. Basically the fix was merging ProxyHelper and ProxyUI.

dunity:
Thanks for pushing those fixes @mtschoen-unity

AndrewTHEManeri:
Vive controllers are not turning transparent for me - expected?

dunity:
The Vive transparency issues have been resolved

AndrewTHEManeri:
Manual Merge successful, performing github merge

PR #359, opened 2017-10-25
Misc improvements

mtschoen-unity:
### Purpose of this PR
Couple of quick fixes

### Testing status

Tested in an empty scene and in Polygon Adventure scene

### Technical risk

Low - Very small amount of changes-intent should be self-evident.

### Comments to reviewers

Creating tooltips when shuffling feedback in locomotion tool was allocating like 30k of garbage. It's not per-frame but it's close, when we're doing tooltip-heavy stuff. Should have been pooled anyway.

I thought the viewer scale visuals were not "pooled" but they are. It's actually activating the object which causes the stutter. I'm going to see how bad it is in .3b6.


AndrewTHEManeri:
still approved

AndrewTHEManeri:

PR #362, opened 2017-11-01
Annotation events.

yipinguo:
Add events for annotation when it is updated or finalized.


mtschoen-unity:
Yiping has been chatting with me about this on Discord. @AndrewTHEManeri brings up a good question, though. Maybe we want to do this as a Functionality Injection interface?

AndrewTHEManeri:
That's what I'm thinking - but I think we'll want it to work backwards from normal.  In that annotation tool would be the one calling the injected method and there would be a third party module responsible for translating those into the callbacks

yipinguo:
@AndrewTHEManeri @mtschoen-unity When the update event is triggered, I serialize the mesh and send it to the other clients. When the finalize event is triggered, I send the mesh and the game objects' new transform.

PR #363, opened 2017-11-07
Feedback Viewing Angle and Affordance Visibility

mtschoen-unity:
### Purpose of this PR
This PR wraps up a couple of Controllers Project improvements:
- Viewing angle detecting and tooltip placement
- Individual affordance visibility control
- Affordances become visible when they have feedback or are animating
- Refactor/simplify/fix implementation

Orientation is detected by comparing the local up/left/right vector to the vector between the camera position and the controller position. Whichever local cardinal direction is closest to the camera vector chooses the FacingDirection. The first AffordanceTooltipPlacement that matches that facing direction for each tooltip is the chosen placement, which provides a source/target transform and an alignment property.
When we set up the proxies, we use a special AffordanceTooltip that has a list of AffordanceTooltipPlacements. I've organized the transform hierarchy with a couple of groups, one for each pair of viewing angles (X,Y,Z for Left/Right, Top/Bottom, Front/Back). Some affordances needed specific left/right targets, but most could use the same target as long as we flip the tooltip based on viewing angle.

I also added a rule to flip the up direction of the tooltip in case you're looking at the controller upside-down. With the two new flipping rules (flip Y if the Z is facing toward the camera, flip Z if the Y is facing down or toward the camera), you shouldn't ever see a tooltip backwards or upside-down. This lets us re-use a lot of tooltip targets which would normally just need to be flipped.

There is some smoothing code in the previous implementation which isn't replicated here. I'm not sure it is necessary, but I've brought up the issue with @dunity and we'll decide if it is necessary to bring back for this PR.

### Testing status
Tested in an empty scene and in Polygon Adventure scene. Tested feedback from locomotion, Selection, and Annotation/CreatePrimitive tools.

### Technical risk
Medium - This touches pretty much everything in the feedback system. Thankfully, there were already some bugs that needed fixing.

### Comments to reviewers
I put ProxyAnimator in a namespace, so view with ?w=1

Two major choices motivated the majority of the code changes here. One was to merge ProxyUI and ProxyHelper, which eliminated the need for any of that special setup code, and the other was to create the AffordanceData class and associated collection to contain all per-affordance runtime state information. At this point, it contains visibility information, but may contain animation or interaction states in the future. Using AffordanceData eliminates the need to clone the AffordanceMap and its definitions. We now treat the serialized data as immutable, and only reference it in runtime types.

Something I didn't get to in this PR, but would like to fix in the future is to reuse or make constant as many feedback requests as possible, like k_ShakeFeedbackRequest. We currently create/destroy them every time they are prevented, which is a little wasteful GC-wise.


Scripts/Helpers/ProxyNode.cs
```diff
@@ -0,0 +1,626 @@
...
+            {
+                kvp.Value.OnDestroy();
+            }
+
+            m_BodyData.OnDestroy();
+        }
+
+        FacingDirection GetFacingDirection(Vector3 cameraPosition)
+        {
+            var transform = m_NaturalOrientation;
```
dunity: Why cache off m_NaturalOrientation, as opposed to just using m_NaturalOrientation?

Scripts/Modules/TooltipModule/TooltipModule.cs
```diff
@@ -39,6 +42,15 @@ class TooltipData
...
             public bool persistent;
             public float duration;
+            public ITooltipPlacement placement;
+
+            public Transform GetTooltipTarget(ITooltip tooltip)
+            {
+                if (placement != null)
+                    return placement.tooltipTarget;
+
+                return ((MonoBehaviour)tooltip).transform;
```
dunity: Is it safe to always assume this cast, before polling for transform?  What if something other than a MonoBehaviour implements ITooltip?

Scripts/Proxies/Vive/ViveProxy.cs
```diff
@@ -39,18 +39,9 @@ public override void Awake()
             var proxyHelper = m_LeftHand.GetComponent<ViveProxyHelper>();
             if (proxyHelper)
             {
-                foreach (var tooltip in proxyHelper.leftTooltips)
+                foreach (var @override in proxyHelper.leftPlacementOverrides)
```
dunity: Not that it matters much, but why not just use a keyword that isn't reserved, like "placementOverride"?  Are @ literals acceptable in the conventions (or is there no verdict on their usage)? @AndrewTHEManeri thoughts?  (Not that it matters at the compiler level)

Scripts/Utilities/MaterialUtils.cs
```diff
@@ -107,6 +97,16 @@ public static Color HueShift(Color color, float shift)
             hsv.x = Mathf.Repeat(hsv.x + shift, 1f);
             return Color.HSVToRGB(hsv.x, hsv.y, hsv.z);
         }
+
+        public static void AddMaterial(this Renderer @this, UnityMaterial material)
```
dunity: Add summary

Scripts/Proxies/Vive/ViveProxy.cs
mtschoen-unity: Yeah, I've started using these mores just because it's all the same with autocomplete. placementOverride is fine here if we want to avoid @s

Scripts/Helpers/ProxyNode.cs
mtschoen-unity: Ah, good catch. Laziness--I was using the builtin transform before. I'll clean this up :)

Scripts/Helpers/ProxyNode.cs
```diff
@@ -0,0 +1,626 @@
...
+        [Tooltip("Affordance objects that store transform, renderer, and tooltip references")]
+        [SerializeField]
+        Affordance[] m_Affordances;
+
+        readonly Dictionary<VRControl, AffordanceData> m_AffordanceData = new Dictionary<VRControl, AffordanceData>();
+        AffordanceData m_BodyData;
+
+        static readonly ProxyFeedbackRequest k_ShakeFeedbackRequest = new ProxyFeedbackRequest { showBody = true };
+
+        FacingDirection m_FacingDirection = FacingDirection.Back;
```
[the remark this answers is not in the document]
mtschoen-unity: Good call

Scripts/Helpers/ProxyNode.cs
```diff
@@ -0,0 +1,626 @@
...
+        public int priority;
+        public VRControl control;
+        public Node node;
+        public string tooltipText;
+        public bool suppressExisting;
+        public bool showBody;
+        public float druation = 5f;
+    }
+
+    class ProxyNode : MonoBehaviour, ISetTooltipVisibility, ISetHighlight, IConnectInterfaces
```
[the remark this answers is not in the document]
mtschoen-unity: Should this class be public/documented? I wonder if we want to support it directly, or just encourage people to use it by duplicating our proxies or just make their own ProxyBase and ProxyNode.

Scripts/Helpers/ProxyNode.cs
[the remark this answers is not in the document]
mtschoen-unity: Ah right... I remember now. The default for an enum is still 0 regardless of whether the enum implements a zero value.

Scripts/Modules/TooltipModule/TooltipModule.cs
mtschoen-unity: This has come up once before. The problem is that otherwise we have literally no way of knowing where to position the tooltip. We could just go ahead and implement the default view-space tooltip now, but otherwise, a tooltip without placement that is not a MonoBehaviour is an invalid case and should throw an error.

Scripts/Proxies/Vive/ViveProxy.cs
AndrewTHEManeri: I'd avoid the @s - there is a whole world of variable names we can use out there

Scripts/Helpers/ProxyNode.cs
```diff
@@ -0,0 +1,619 @@
...
+            foreach (var request in m_FeedbackRequests)
+            {
+                if (request.caller == caller)
+                    k_FeedbackRequestsCopy.Add(request);
+            }
+
+            foreach (var request in k_FeedbackRequestsCopy)
+            {
+                RemoveFeedbackRequest(request);
+            }
```
AndrewTHEManeri: you might want another clear of k_FeedbackRequestsCopy here - which I know is redundant but for GC purposes it just makes sure certain assets aren't referenced

mtschoen-unity:
I've updated ProxyNode to keep track of visibility by renderer/material, rather than by affordance. The issue came up when testing the thumbstick animator feedback. Even if you kept one axis from being zero, if the other axis is zero for a frame, it will try to hide the thumbstick. We were still keeping two visibility states and current color values for the same material, and those were "fighting" over the visibility state.

We now organize VisibilityState objects by cloned material, and choose the one which will be hidden last to govern the transition (or the body visibility if it is visible).

One of the use cases that we considered (aside from thumbstick/thumbpad) would be a hand model. If I want to highlight individual fingers on a single mesh, we need to specify which material refers to which finger/control. Thus, I added an optional override material to specify which of the renderer's materials the affordance relates to. If it relates to more than one material, authors can add more than one affordance definition for a given control and the materials will both "catch" visibility states.

I haven't found any bugs with the current state, but haven't tested extensively.

AndrewTHEManeri:

PR #365, opened 2017-11-09
Add a setting to EditorVRContext for VR supersampling

mtschoen-unity:
### Purpose of this PR
Make use of XRSettings.eyeTextureResolutionScale which was added in 2017.2

### Testing status
Tested in an empty scene and in Polygon Adventure scene. Performance is better when values are < 0, and worse when > 0. Text is much more readable at 2x.

### Technical risk
Low - Just sets a property based on a serialized value

### Comments to reviewers
One downside here is that the setting will be tracked by version control. We could make this into a preference, but I didn't think it would be something you change very often.


AndrewTHEManeri:
It's fun setting the renderscale to .1 and pretending I am in an old school adventure game.

PR #369, opened 2017-11-13
Persistent Undo/Redo on radial menu

mtschoen-unity:
### Purpose of this PR
Give the user constant access to Undo/Redo

### Testing status
Tested in an empty scene and in Polygon Adventure scene. Tested selecting/deselecting/direct selecting to test main radial menu

### Technical risk
Medium - This makes changes to how the radial menu is displayed/hidden. Also adds support for changing the actions displayed on the radial menu, which was part of the original spec but never made it.

### Comments to reviewers
I had to make some changes to animations and transitions because the way it worked at first was to hide/show the entire radial menu every time. While we were at it, @dunity and I decided to rotate the starting point of the radial menu down to where the menu activator is.

Also, this branch has been out for quite a while and I'm not sure if there are any nasty interactions with latest dev that I missed. Please test thoroughly.


dunity:
Overall looks good.  Just a few minor changes suggested.

dunity:
There is a visual issue that occurs when opening the main menu, via selecting the main menu button.  After selecting the main menu button, the ray (for me) is almost always hovering one of the two buttons (undo/redo), though the buttons are not being highlighted, and are still in their un-hovered state.  The buttons should be in the hovered state when such an event occurs.

Menus/RadialMenu/Scripts/RadialMenuSlot.cs
```diff
@@ -476,7 +470,8 @@ IEnumerator AnimateSemiTransparent(bool makeSemiTransparent)
             var transitionAmount = Time.deltaTime;
             var positionWait = (orderIndex + 4) * 0.25f; // pad the order index for a faster start to the transition
             var currentScale = transform.localScale;
-            var targetScale = Vector3.one;
+            var semiTransparentTargetScale = new Vector3(0.9f, 0.15f, 0.9f);
```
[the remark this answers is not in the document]
mtschoen-unity: Hm. This might be a merge fail, because I don't remember writing this. I'll see if this is intentional

Scripts/Core/EditorVR.Menus.cs
```diff
@@ -430,23 +430,47 @@ internal static void UpdateAlternateMenuOnSelectionChanged(Transform rayOrigin)
                 if (rayOrigin == null)
                     return;
 
-                SetAlternateMenuVisibility(rayOrigin, Selection.gameObjects.Length > 0);
+                UpdateAlternateMenuActions();
```
jono-unity: Seems this would be clearer as `DefaultAlternateMenuActions()` or such.

Scripts/Modules/ActionsModule.cs
```diff
@@ -9,10 +9,27 @@ namespace UnityEditor.Experimental.EditorVR.Modules
 {
     sealed class ActionsModule : MonoBehaviour, IConnectInterfaces
     {
-        public List<ActionMenuData> menuActions { get { return m_MenuActions; } }
+        public List<ActionMenuData> menuActions
```
jono-unity: Seems like this should be `defaultMenuActions`

AndrewTHEManeri:
@dunity Re: Hovered button bug.  Yes, I agree they should be hovered there.  In general, if clicking would make the button do something, it should show the highlighted state

mtschoen-unity:
I'm wondering about this branch's fate. @jono-unity is working on an alternate undo/redo UI which will make the persistent undo redundant. But, I would like to keep the functionality of adding/removing actions from the Radial menu, as opposed to simply keeping the current implementation.

I propose that we close this PR and leave the branch open until after the dust settles, then graft the Radial changes minus the always-undo part.

AndrewTHEManeri:
That sounds reasonable to me.  Make it so

mtschoen-unity:
And then there were 9.

PR #371, opened 2017-11-13
fix "click stick to deselect" in radial menu

stella3d:
does what it says - fixes https://favro.com/card/c564ede4ed3337f7b17986b6/Uni-31181

tested manually, and all radial actions seem to work as expected, nothing else i saw that this breaks.

### DIFF
--- a/Menus/RadialMenu/RadialMenu.cs
+++ b/Menus/RadialMenu/RadialMenu.cs
@@ -128,8 +128,7 @@ namespace UnityEditor.Experimental.EditorVR.Menus
             }
             else
             {
-                m_RadialMenuUI.buttonInputDirection = Vector3.zero;
-                return;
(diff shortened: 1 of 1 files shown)


Menus/RadialMenu/RadialMenu.cs
```diff
@@ -128,8 +128,7 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
             }
             else
             {
-                m_RadialMenuUI.buttonInputDirection = Vector3.zero;
-                return;
+                m_RadialMenuUI.buttonInputDirection = Vector2.zero;
```
AndrewTHEManeri: This was seriously it?

mtschoen-unity: Wait, was this the Vector2 change, or the removing of the return that fixed it?

PR #376, opened 2017-11-14
fix toolflow ray selection

stella3d:
the gradient buttons were non-interactable in the prefab, and when creating a new button we did not set interactable to default to true.   This changes both of those things, so that a gradient button is interactable until it's disabled.


AndrewTHEManeri:

PR #380, opened 2017-11-14
fix project view scrollback when there is only one row

stella3d:
there were two problems:

`listHeight` would be 0 when there were less items than items per row.  this doesn't seem like how `Mathf.CielToInt` should work, but it does.  Now it's clamped to multiply by a minimum of one row.

we were scrolling back once we the negative scroll offset was greater than or equal to the list height, but the first row leaves the screen when scroll offset is < 0.   

i tried having the scroll back point adjust itself based on these scenarios, but those implementations didn't prove as immediately clean as just handling it as a separate case.

### DIFF
--- a/Workspaces/ProjectWorkspace/Scripts/AssetGridViewController.cs
+++ b/Workspaces/ProjectWorkspace/Scripts/AssetGridViewController.cs
@@ -50,7 +50,8 @@ namespace UnityEditor.Experimental.EditorVR.Workspaces
                 if (m_NumPerRow == 0)
                     return 0;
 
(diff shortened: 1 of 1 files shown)


AndrewTHEManeri:
This works but I'm not a fan of how scrolling down just stops smoothly with 1 row, but scrolling up kind of 'respawns' the items over and over again.  Is that the snap back?

@dunity - thoughts?

stella3d:
@AndrewTHEManeri  that is the snapback - that's how things get snapped back when scrolling the other way, though.

PR #381, opened 2017-11-15
Fix nullref in ProjectFolderModule.CreateAssetData.

jono-unity:
AssetDatabase.GetMainAssetTypeAtPath returns null on some types of assets, and was throwing nullrefs. This change leaves typeName empty in that case.


AndrewTHEManeri:

PR #382, opened 2017-11-15
Block Select

### LINKED ISSUE #351: Prevent ray-based selection of player(head)
Allow only direct-selection of the user's head; don't allow the ray to interact with the player.  Aiming the input-device(controller) at the player's head activates feedback display.

mtschoen-unity:
### Purpose of this PR
Allow selection of all objects within a volume (currently cube/sphere)

### Testing status
Tested in an empty scene and in Polygon Adventure scene. Tested with multiselect on/off, and with existing selections, etc.

### Technical risk
Medium - Selection is a pretty central feature, and this has the potential to modify/break normal selection behavior

### Comments to reviewers
The scene view outline (not ours) has a heavy cost on selection based on how many objects you are selecting and how complex they are. I believe that this is due to some CPU work to soften edges or something. If you disable it, you don't get the big hitch when selecting, but you can't see what you've selected once it's selected.

For the SIGGRAPH demo, because the Presentation Camera also doesn't show scene view selections, I modded the HighlightModule to always highlight selected objects. I figured this would be a performance concern, but it did not seem to be an issue. I could do this here, and recommend to users that they turn off scene highlight.

The introduction of the Standard Ignore List will help with #351


Scripts/Utilities/IntersectionUtils.cs
```diff
@@ -183,6 +183,94 @@ public static bool TestRay(MeshCollider collisionTester, Transform obj, Ray ray)
...
+                return false;
+
+            //Because our sphere check cannot be non-uniformly scaled, transform the test object instead
+            var testerTransform = collisionTester.transform;
+            testerTransform.position = obj.position;
+            testerTransform.rotation = obj.rotation;
+
+            //Negative scales cause mesh read errors
+            var objScale = obj.lossyScale;
+            UndoInverseScale(ref objScale.x, ref center, obj);
```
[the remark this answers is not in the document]
mtschoen-unity: Yeah this was a really tricky problem to solve, so there are probably some improvements that can be made. What do you mean by "the whole vector at once?" As in, if I have two/three inversions I don't have to rotate? I could probably add special cases for that but it would be a _lot_ more code.

mtschoen-unity:
Are you talking about the hitch after the selection is made? I'll look at the profiler and see if anything jumps out as a quick fix. Looks like I need to merge dev again, too.

mtschoen-unity:
There were a couple of dumb things happening with big selections:
- SelectObject did operations on the whole selection, and was called individually for every object in a block select. I added a SelectObjects to do as much as possible in a single pass. Using a hashset to manipulate the current selection was also a big win.
- TransformTool setting positions on objects caused them to come up in SpatialHashModule.UpdateDynamicObjects, which for a large selection was costing us a lot of time

I still see a slight hitch when doing a block selection which comes down to getting the bounds on such a large selection in the TransformTool to decide where to place the manipulator if it's in center mode. Granted, we don't need to do this for non-center pivot mode, but we should try to optimize this in a later PR.

AndrewTHEManeri:
That has helped *greatly* - thanks!

Scripts/Core/Interfaces/IUsesPointer.cs
```diff
@@ -0,0 +1,26 @@
+﻿#if UNITY_EDITOR
+using System;
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR
+{
+    interface IUsesPointer
```
dunity: Add summary

Scripts/Core/Interfaces/IUsesPointer.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+    }
+
+    static class IUsesPointerMethods
+    {
+        internal static Func<Transform, float> getPointerLength { get; set; }
+
+        public static float GetPointerLength(this IUsesPointer obj, Transform rayOrigin)
```
dunity: Add summary

Scripts/Core/Interfaces/IUsesPointer.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            return getPointerLength(rayOrigin);
+        }
+
+        public static Vector3 GetPointerPosition(this IUsesPointer obj, Transform rayOrigin)
```
dunity: Add summary

Scripts/Interfaces/FunctionalityInjection/IStandardIgnoreList.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+
+#if UNITY_EDITOR
```
dunity: Move if to top of class

Scripts/Modules/SelectionModule/SelectionModule.cs
```diff
@@ -13,13 +14,16 @@ sealed class SelectionModule : MonoBehaviour, IUsesGameObjectLocking, ISelection
...
         GameObject m_CurrentGroupRoot;
-        readonly List<Object> m_SelectedObjects = new List<Object>(); // Keep the list to avoid allocations--we do not use it to maintain state
 
         public Func<GameObject, GameObject> getGroupRoot { private get; set; }
         public Func<GameObject, bool> overrideSelectObject { private get; set; }
 
         public event Action<Transform> selected;
 
+        // Local method use only -- created here to reduce garbage collection
+        static readonly HashSet<Object> m_SelectedObjects = new HashSet<Object>();
```
dunity: k_SelectedObjects

mtschoen-unity:
@dunity I've responded to your comments.

@AndrewTHEManeri something to check out is the fix in 3e2b771. I couldn't reproduce it in any other branch, and I think it (somehow) has to do with consuming the trigger. In any case, the issue seems to be independent of Block Select but was also fixed by removing  the input consumption when block selecting in SelectionTool. Go figure.

AndrewTHEManeri:

PR #387, opened 2017-11-16
snapping settings "back" tooltip

stella3d:
This adds a `TooltipData` component class for easily adding tooltips to existing objects, and uses that component to add a tooltip to the snapping menu back button. 

i didn't see a clear standard for positioning of tooltips, so i just set it to something that looked good to me.  if there's a standard, i'd like to make it follow that.   for instance, i see with the main menu tooltip that it's quite a distance from the actual button, but that felt awkward for the top of the main menu.
snapsettingstooltip

### DIFF
--- a/Scripts/UI/TooltipData.cs
+++ b/Scripts/UI/TooltipData.cs
@@ -1,0 +1,44 @@
+#if UNITY_EDITOR
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR.UI
+{
+    /// <summary>
+    /// Place on an object that recieves UI events to set a tooltip on it
+    /// </summary>
+    class TooltipData : MonoBehaviour, ITooltip, ITooltipPlacement
+    {
+        [SerializeField]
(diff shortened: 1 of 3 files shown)


dunity:
At first glance, I'd rephrase the class comment/summary.  It could be a bit unclear for someone getting into EVR.

Scripts/UI/TooltipData.cs
```diff
@@ -0,0 +1,44 @@
+#if UNITY_EDITOR
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR.UI
+{
+    /// <summary>
+    /// Place on an object that recieves UI events to set a tooltip on it
```
dunity: Correct spelling : "receives"

dunity:
I just pushed a fix for the typo I cited in my review.

dunity:
In addition, I raised the tooltip location a bit, making it obscure the header text less at certain angles.

Issue #383, opened 2017-11-16
MainMenu sub-face/custom buttons lacking haptics


dunity:
Custom main menu sub-face-buttons (snapping, locomotion, etc), are lacking the standard haptic(s) implementation.  These buttons should implement the same haptics for hover/etc as that of the standard main menu buttons (tools, workspaces, etc).

stella3d:
i'll grab this one, i haven't implemented any haptics yet and would love to

PR #392, opened 2017-11-17
sub / settings menu haptics update

stella3d:
This adds haptic feedback to all the Toggles that are on the main menu that did not have them. 

In the process, i made a base class for the menu toggles and buttons.  The events aren't in the base class because working with inherited events proved to not work well.  Toggles have simpler hover actions since they don't generally need / have tooltips.

I used a weaker click pulse, because the longer pulse was too long for toggles, it felt like.  this one feels good to me.

I also had to update the LocomotionTool's settings UI to use TextMeshPro.


stella3d:
this should resolve #383

Menus/MainMenu/Scripts/MainMenuToggle.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        new void Awake()
+        {
+            m_Selectable = m_Toggle;
+            m_OriginalColor = m_Toggle.targetGraphic.color;
+        }
+
+        public void OnRayEnter(RayEventData eventData)
+        {
+#if INCLUDE_TEXT_MESH_PRO
```
AndrewTHEManeri: Don't think this is #if block is needed?

Menus/MainMenu/Scripts/MainMenuSelectable.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        [SerializeField]
+        HapticPulse m_HoverPulse;
+
+#if INCLUDE_TEXT_MESH_PRO
+        [SerializeField]
+        protected TextMeshProUGUI m_Description;
+#endif
+
+#if INCLUDE_TEXT_MESH_PRO
```
mtschoen-unity: Group these fields in the same #ifdef. I know this was my code but I did it wrong :)

Menus/MainMenu/Scripts/MainMenuToggle.cs
mtschoen-unity: Yeah, there are no TMP references in this entire class. Looks like it's left over from before the refactor to MainMenuSelectable

stella3d:
think that should handle updating to merge with latest dev & PR feedback.  removed un-needed if defs.

dunity:
There appear to be conflicts with the SettingsUI prefab & this branch.  Please be careful not to overwrite any recent fixes/changes on that prefab when merging latest development.

mtschoen-unity:
@dunity I merged in dev, updated TMP on the settings UI that needed it, disabled hovers on non-interactable MenuButtons and MenuToggles, and added hovers to Toggles that aren't on. Basically we check interactable, rather than isOn to block hovers.

I also added a MainMenuButton to the "breadcrumb" that says "Snapping" so that it sends hover/click pulses.

Also deleted unused pulses from MainMenuButton.

Approved on my end. Waiting on @dunity's approval to merge do dev.

mtschoen-unity:
Manual merge successful. Performing Github merge.

PR #397, opened 2017-11-18
revert TMP to Text in UI DropDown (Fix Inspector)

stella3d:
This fixes the totally messed up Inspector issue


mtschoen-unity:
Looks like the SDF changes conflict with the ones I checked in in block-select

AndrewTHEManeri:
I'll resolve it when I merge everything

PR #399, opened 2017-11-18
Misc bug fixes on external project

mtschoen-unity:
Implements basic grouping (doesn't save between runs of EVR)
Implements IContainsVRPlayerCompletely to disable selection on objects that surround the player
Fixes "stuck" highlight issue that can happen with direct selection (and is exacerbated by disabled selections by the above)

### DIFF
--- a/Scripts/Utilities/ObjectUtils.cs
+++ b/Scripts/Utilities/ObjectUtils.cs
@@ -1,4 +1,4 @@
-﻿#if UNITY_EDITOR
+#if UNITY_EDITOR
 using System;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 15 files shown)


Scripts/Utilities/ObjectUtils.cs
```diff
@@ -134,6 +129,24 @@ public static Bounds GetBounds(Transform transform)
...
 
+        public static Bounds GetBounds(List<Renderer> renderers)
+        {
+            if (renderers.Count > 0)
+            {
+                var first = renderers[0];
+                var b = new Bounds(first.transform.position, Vector3.zero);
+                foreach (var r in renderers)
+                {
+                    if (r.bounds.size != Vector3.zero)
```
AndrewTHEManeri: why is the zero check needed here?

Tools/SelectionTool/SelectionTool.cs
```diff
@@ -279,7 +279,12 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
 
                     // Can't select this object (it might be locked or static)
                     if (directHoveredObject && !directSelectionCandidate)
+                    {
+                        if (directHoveredObject != null)
```
AndrewTHEManeri: It seems like the first if check does the same thing, is this a Unity-overloaded null operator problem?  (And in that case, can we just move this line up to the first if statement?)

PR #402, opened 2017-11-19
Affordance visibility refactor

mtschoen-unity:
### Purpose of this PR
Refactor visibility transition code and fix Vive "touch pip."

### Testing status
Tested in an empty scene with Vive controllers.

### Technical risk
Low - Only affects ProxyNode and Vive proxy prefab.

### Comments to reviewers
I realized while writing AnimateProperty that I had just gotten lucky in ProxyNode with the math for getting the percentage of the current value toward the goal. This has been fixed, and the code is much more readable now in my opinion.

### DIFF
--- a/Scripts/Proxies/ProxyNode.cs
+++ b/Scripts/Proxies/ProxyNode.cs
@@ -21,6 +21,7 @@ namespace UnityEditor.Experimental.EditorVR.Proxies
                 class VisibilityState
                 {
                     readonly int m_MaterialIndex;
+                    readonly Material m_Material;
                     readonly AffordanceTooltip[] m_Tooltips;
                     readonly AffordanceVisibilityDefinition m_Definition;
 
@@ -48,7 +49,18 @@ namespace UnityEditor.Experimental.EditorVR.Proxies
                     {
                         m_Tooltips = tooltips;
                         m_Definition = definition;
+                        m_Material = material;
                         m_MaterialIndex = Array.IndexOf(renderer.sharedMaterials, material);
+                    }
+
+                    public void SetFloat(float value)
+                    {
+                        m_Material.SetFloat(m_Definition.alphaProperty, value);
+                    }
+
+                    public void SetColor(Color value)
+                    {
+                        m_Material.SetColor(m_Definition.colorProperty, value);
                     }
                 }
 
@@ -56,8 +68,11 @@ namespace UnityEditor.Experimental.EditorVR.Proxies
                 float m_VisibleChangeTime;
                 Color m_OriginalColor;
                 Color m_StartColor;
+                Color m_CurrentColor;
 
                 readonly Dictionary<VRControl, VisibilityState> m_Visibilities = new Dictionary<VRControl, VisibilityState>();
+                readonly Dictionary<KeyValuePair<Material, string>, Action<float>> m_SetFloats = new Dictionary<KeyValuePair<Material, string>, Action<float>>();
+                readonly Dictionary<KeyValuePair<Material, string>, Action<Color>> m_SetColors = new Dictionary<KeyValuePair<Material, string>, Action<Color>>();
 
                 public void AddAffordance(Material material, VRControl control, Renderer renderer,
                     AffordanceTooltip[] tooltips, AffordanceVisibilityDefinition definition)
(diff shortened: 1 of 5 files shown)


Scripts/Proxies/ProxyNode.cs
```diff
@@ -106,76 +124,58 @@ public VisibilityState(Renderer renderer, AffordanceTooltip[] tooltips, Affordan
...
                     var fadeDuration = visible ? fadeInDuration : fadeOutDuration;
-
                     switch (definition.visibilityType)
                     {
                         case VisibilityControlType.AlphaProperty:
-                            var alphaProperty = definition.alphaProperty;
-                            if (visible != m_WasVisible)
-                                m_StartColor = material.GetFloat(alphaProperty) * Color.white;
+                            Action<float> setFloat = null;
+                            if (visibilityState != null)
```
AndrewTHEManeri: Would we be better served just making a visibility state here rather than a setfloat function that is essentially the same thing as visibilitystate's setfloat?

Scripts/Proxies/ProxyNode.cs
```diff
@@ -106,76 +124,58 @@ public VisibilityState(Renderer renderer, AffordanceTooltip[] tooltips, Affordan
...
-                            var colorProperty = definition.colorProperty;
-                            if (visible != m_WasVisible)
-                                m_StartColor = material.GetColor(colorProperty);
+                            Action<Color> setColor = null;
+                            if (visibilityState != null)
+                                setColor = visibilityState.SetColor;
 
-                            var targetColor = visible ? m_OriginalColor : definition.hiddenColor;
-                            if (m_StartColor != targetColor)
+                            if (setColor == null)
```
AndrewTHEManeri: same as above comment

Scripts/Proxies/Vive/ViveProxy.cs
```diff
@@ -73,9 +73,10 @@ static void PostAnimate(Affordance[] affordances, AffordanceDefinition[] afforda
                         }
                         break;
                     case VRInputDevice.VRControl.Analog0:
+                        // TODO: Support multiple transforms per control
```
AndrewTHEManeri: Captured in https://github.com/Unity-Technologies/EditorVR/issues/403

Scripts/Utilities/TransitionUtils.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            var timeDiff = time - changeTime;
+            var source = state ? startValue : endValue;
+            var target = state ? endValue : startValue;
+            if (!approximately(property, target))
+            {
+                var duration = (1 - getPercentage(propertyStart, target, source)) * totalDuration;
+                var smoothedAmount = MathUtilsExt.SmoothInOutLerpFloat(timeDiff / duration);
+                if (smoothedAmount > 1)
```
AndrewTHEManeri: '>=?'

Scripts/Utilities/TransitionUtils.cs
```diff
@@ -0,0 +1,71 @@
...
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR.Utilities
+{
+    /// <summary>
+    /// Transition utilities
+    /// </summary>
+    public static class TransitionUtils
+    {
+        public static void AnimateProperty<T>(float time, bool state, ref bool lastState, ref float changeTime,
```
AndrewTHEManeri: Do we want some standard for organizing args?  Like normal, ref, out, function (or some other order?)

Scripts/Utilities/TransitionUtils.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public static float GetPercentage(float current, float target, float source)
+        {
+            return (current - source) / (target - source);
+        }
+
+        public static float GetPercentage(Color currentColor, Color targetColor, Color sourceColor)
+        {
+            var current = currentColor.grayscale * currentColor.a;
```
AndrewTHEManeri: Any reference for what conversion to grayscale Unity uses? Average? Max?  Min?

Scripts/Utilities/TransitionUtils.cs
dunity: Ideally, yeah.  IMO there should be a standard for the core args.

Scripts/Utilities/TransitionUtils.cs
dunity: I assumed it was using the traditional dot-based approach.

Scripts/Utilities/TransitionUtils.cs
```diff
@@ -0,0 +1,71 @@
+﻿#if UNITY_EDITOR
+using System;
+using System.Reflection;
```
dunity: Remove: Reflection, Modules, and UI usings

Scripts/Utilities/TransitionUtils.cs
dunity: Add summaries to functions

Scripts/Utilities/TransitionUtils.cs
dunity: Looks like that is the approach used for the hard-set, so the value meets, or matches the target as closely as possible (depending on delta) before the hard-set.

mtschoen-unity: Yes. The == 1 is handled by the approximately check above. If we do miraculously have a value of 1, we'll get the same result from the lerp below.

Scripts/Utilities/TransitionUtils.cs
mtschoen-unity: It's actually kind of a funky average that has different weighting for different colors (I checked the source). I just needed a way to "collapse" the color to a single value. If we want to just do a average, that's fine. Min/max won't work because you will end up stopping prematurely.

Scripts/Proxies/ProxyNode.cs
mtschoen-unity: I went back and forth on this. Ultimately decided the overhead of the unused variables wasn't worth it, but it would be cleaner to read. I'll give this a shot.

Scripts/Utilities/TransitionUtils.cs
mtschoen-unity: I'm not sure I agree that ref/out/standard is the criteria for argument order. Obviously default parameters have to go last as required, but otherwise I try to group them conceptually. In this case, I actually switched back and forth between ref as I was writing the method, and I can imagine that future refactors might turn certain arguments into or out of being ref.

AndrewTHEManeri:

PR #417, opened 2017-11-21
Fix bug where player head is selectable

### LINKED ISSUE #351: Prevent ray-based selection of player(head)

mtschoen-unity:
### Purpose of this PR

Fix an issue where "Select" feedback shows up when you point at your head. Fixes #351 

### Testing status

Tested in the latest EXR in Polygon Adventure Scene. Pointing at my head with feedback always-on (set maxpresentations to 200000) does not show Select feedback. Can still re-position head in miniworld

### Technical risk

Low - Mostly refactor code. Actual fix resides in SelectionTool

### Notes to reviewers

This looks a lot scarier than it is. I wanted to avoid having to GetComponent for the renderer in SelectionTool, and I think it makes sense in general for the ignore lists to be GameObjects. Previously, we were only ever dealing with renderers, so it made sense to use renderers. I don't think we will have the opposite issue (having to get renderers from the objects in the ignorelist) but if so, maybe we can hold a tuple of GameObject/Renderer.

### DIFF
--- a/Scripts/Core/EditorVR.Viewer.cs
+++ b/Scripts/Core/EditorVR.Viewer.cs
@@ -58,7 +58,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
             PlayerBody m_PlayerBody;
             float m_OriginalNearClipPlane;
             float m_OriginalFarClipPlane;
-            readonly List<Renderer> m_VRPlayerObjects = new List<Renderer>();
+            readonly List<GameObject> m_VRPlayerObjects = new List<GameObject>();
 
             readonly Preferences m_Preferences = new Preferences();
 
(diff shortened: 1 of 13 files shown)


Scripts/Core/EditorVR.Viewer.cs
```diff
@@ -197,7 +197,12 @@ internal void AddPlayerModel()
                 m_PlayerBody = ObjectUtils.Instantiate(evr.m_PlayerModelPrefab, CameraUtils.GetMainCamera().transform, false).GetComponent<PlayerBody>();
                 var renderer = m_PlayerBody.GetComponent<Renderer>();
                 evr.GetModule<SpatialHashModule>().spatialHash.AddObject(renderer, renderer.bounds);
-                renderer.GetComponentsInChildren(true, m_VRPlayerObjects);
+                var playerObjects = m_PlayerBody.GetComponentsInChildren<Renderer>(true);
```
stella3d: really these are `playerRenderers`, you get objects attached to them, but not a big deal

PR #431, opened 2017-11-27
EditorVR camera better represents what is seen in the game view

dunity:
### Purpose of this PR

This improvement allows for the EditorVR VRView camera to better represent what is seen in the game view.

Add support for copying the scene's MainCamera settings (clear flags, background, etc), if one exists, and the (EditorVR)context's m_CopySceneCameraSettings inspector bool/toggle is enabled.

### Testing status

Tested in various scenes and camera settings scenarios.  Tested with and without a MainCamera in the scene.

### Technical risk

Medium - The majority of the changes are relegated to EditorVRContext, EditingContextManager, and VRView.  These are core EVR elements; some of their fields/properties have been made static, in order to allow for this functionality to occur as/when needed.  Care has been taken to avoid conflicting with the previous AddComponent<Camera> approach.  Disabling the m_CopySceneCameraSettings bool on the EditorVR context will disable this functionality, and create the camera in the manner that was previously occurring.

### Notes

I've also added initial support for the Unity PostProcessing(v2) stack in the following branch: "improvements/dylanu/scene-camera-fx-settings".  There is more work to be done in that branch before PR.  This branch/PR acts as the foundation for that work.


Scripts/Core/Contexts/EditorVRContext.cs
```diff
@@ -7,30 +7,47 @@
 
 namespace UnityEditor.Experimental.EditorVR.Core
 {
+    public struct ContextSettings
```
AndrewTHEManeri: This should be in a separate file so that IEditingContext is not dependent on EditorVRContext

Scripts/Interfaces/Entity/IEditingContext.cs
```diff
@@ -12,6 +14,11 @@ public interface IEditingContext
         /// </summary>
         string name { get; }
 
+        /// <summary>
+        /// Settings denoting that the scene camera's (component_ values should be cloned on the XR runtime camera
+        /// </summary>
+        ContextSettings contextSettings { get; }
```
AndrewTHEManeri: Why this context settings class instead of individual settings items being exposed as interface members/requirements?

dunity:
@AndrewTHEManeri These changes intentionally allow for copying of any camera settings that aren't replaced by those that are defined in VRView OnEnable(), including forward/deferred/etc.  VRView still sets/replaces values for CameraType/OcclusionCulling/ClipPlanes.  Ideally, we could also support an EVR "CameraCopy" component, that users could set on their desired camera.  Such a component would have toggles allowing for individual copying of specific params, etc.

Scripts/Interfaces/Entity/IEditingContext.cs
dunity: I was initially trying to avoid adding the copySceneCameraSettings param to IEditingContext.  The initial number of settings were more numerous when accounting for PostFX settings (coming in a latter PR).  Though after I minimized the implementation, it now makes sense to merge it back into IEditingContext, and remove the ContextSettings struct (for now).

dunity:
ContextSettings struct removed from IEditingContext & EditorVRContext. copySceneCameraSettings property added to IEditingContext.  This is used in VRView OnEnable() when setting up the VRView camera.

dunity:
@AndrewTHEManeri  your suggested changes have been applied (as I understood them).

Scripts/Utilities/ObjectUtils.cs
```diff
@@ -200,6 +200,34 @@ public static Component AddComponent(Type type, GameObject go)
             return component;
         }
 
+        public static T CopyComponent<T>(T sourceComponent, GameObject targetGameObject) where T : Component
```
mtschoen-unity: Check out EditorUtility.CopySerialized. You should be able to just AddComponent, and then CopySerialized(source, target)

Scripts/Core/Contexts/EditorVRContext.cs
```diff
@@ -13,24 +13,31 @@ class EditorVRContext : ScriptableObject, IEditingContext
...
-
             XRSettings.eyeTextureResolutionScale = m_RenderScale;
         }
 
         public void Dispose()
         {
-            m_Instance.Shutdown(); // Give a chance for dependent systems (e.g. serialization) to shut-down before destroying
-            ObjectUtils.Destroy(m_Instance.gameObject);
-            m_Instance = null;
+            if (m_Instance)
```
mtschoen-unity: Why the null check?

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -24,26 +24,26 @@ sealed class EditingContextManager : MonoBehaviour
...
+        static List<IEditingContext> s_AvailableContexts;
+        static EditingContextManagerSettings s_Settings;
+        static UnityObject s_DefaultContext;
+        static IEditingContext s_CurrentContext;
 
-        EditingContextManagerSettings m_Settings = null;
-
-        List<IEditingContext> m_AvailableContexts;
-        string[] m_ContextNames = null;
+        string[] m_ContextNames;
```
mtschoen-unity: Making all of this static is a little worrying to me, but I haven't found any issues in testing.

dunity: There's a need to access some context data before instantiation occurs.  I ran a ton of tests too, and couldn't find any issues.  If anyone wants to suggest, or craft a better solution, I'm open to any input/changes.

Scripts/Core/Contexts/EditorVRContext.cs
dunity: I ran into a case that produced an exception when calling Dispose() before the EXR instance is created in IEditingContext's Setup() function.  I've just pushed an alternate fix that validates that an instance has been created, via the instanceExists bool prop in IEditingContext, before attempting to dispose in EditingContextManager's SetEditingContext() function.

Scripts/Utilities/ObjectUtils.cs
dunity: good call; I'm refactoring the method now

dunity:
Suggested changes pushed.  Please give it another look when you have the time @mtschoen-unity , thanks :-)

mtschoen-unity:
Forgot to mention in my commit message that the last one also fixes an issue with the Editing Context popup in VRView if there is >1 context available. It was broken since the GUILaout->GUI refactor.

@AndrewTHEManeri I'm done making changes and this should be good to go. Still waiting on your approval.

jono-unity:
@dunity here's some funky behavior: it looks like if 'Allow HDR' is false on a camera, the presentation camera gets squished. I'm seeing this in an empty (just camera and light) scene and in the Poly Pirates main scene, and not on dev.  In this vid, I'm toggling 'presentation camera': https://youtu.be/Zpk_kMgKGCU

mtschoen-unity:
@jono--can confirm. Away it goes, then. I'll force HDR on like I did with the other settings that break things.

mtschoen-unity:
@stella3d this feature is a perfect example for your AI monkey script. It could run EXR with all permutations of camera settings and record failures. :+1:

mtschoen-unity:
Manual merge successful. Performing Github merge

PR #432, opened 2017-11-28
deselecting now hides the keyboard if it's open

### LINKED ISSUE #404: Numeric keyboard should be hidden, if it is still visible, when lacking context.
The numeric keyboard, should hide when an object is deselected, and the inspector isn't displaying any object data.

In any other context, in which the numeric keyboard's input won't be valid, it should also be hidden.

stella3d:
fixes #404.  
The keyboard stays open when you select another object to inspect, but closes if you deselect.

### DIFF
--- a/Scripts/UI/InputField.cs
+++ b/Scripts/UI/InputField.cs
@@ -102,6 +102,12 @@ namespace UnityEditor.Experimental.EditorVR.UI
             // Don't do base functionality
         }
 
+        protected override void OnDisable()
+        {
+            if (m_KeyboardOpen && Selection.activeObject == null)
+                CloseKeyboard(true);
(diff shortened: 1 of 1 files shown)


Scripts/UI/InputField.cs
```diff
@@ -47,7 +48,7 @@ public class OnChangeEvent : UnityEvent<string>
         [SerializeField]
         private int m_CharacterLimit = 10;
 
-        private bool m_KeyboardOpen;
+        protected bool m_KeyboardOpen;
```
mtschoen-unity: This protected does not seem to be necessary. I'm removing it.

mtschoen-unity:

PR #434, opened 2017-11-28
force Main Menu animation to set full blend weight at end

### LINKED ISSUE #433: Main menu height not expanding fully on-reveal/show
When opening the main menu, the "ShinkHeight" blendshape value on the "MenuFrame" object is not being fully set to 0(zero).  The MenuFrame object can be found in the MainMenu prefab.   The animated reveal should end with a hard-set of the ShinkHeight blendshape value to zero.

stella3d:
fixes #433 

`currentDuration - Time.deltaTime` is to allow the lerp to run for another frame & get closer to the target weight, based on suggestion from @dunity

### DIFF
--- a/Menus/MainMenu/Scripts/MainMenuUI.cs
+++ b/Menus/MainMenu/Scripts/MainMenuUI.cs
@@ -490,8 +490,6 @@ namespace UnityEditor.Experimental.EditorVR.Menus
                 yield return null;
             }
 
-            m_VisibilityState = visibilityState;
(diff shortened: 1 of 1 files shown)


mtschoen-unity:
Inspected MenuFrame and confirmed. Blend weights all good. SO GLAD THIS IS FIXED!

mtschoen-unity:
Manual merge succeeded. Performing GitHub merge.

PR #437, opened 2017-11-29
Fix Unlock All button on LockedObjectsWorkspace

mtschoen-unity:
Tiny little change:
- HierarchyData is now represented as a single root object with children, so we need to use the children list now when unlocking all.
- LockedObjectsWorkspace was not updated to the new style of Functionality Injection, so we don't need the delegate properties anymore


mtschoen-unity:
Manual merge successful. Performing GitHub Merge

PR #438, opened 2017-11-30
Controller tooltips

mtschoen-unity:
Prefabs only. Pull this and check out the multiselect and affordance tooltip positions with radial menu open


mtschoen-unity:

Issue #441, opened 2017-11-30
Presentation camera shouldn't take into account the previous-disabled rotation when enabled


dunity:
Currently, when enabling the presentation camera (VRSmoothCamera), there can be a moment of very quick rotation that looks distracting.  When apparent, this results in an odd looking, untended rotation (based on the rotation delta from the previous m_Rotation that was lerped in LateUpdate()).  The roll rotation correction between the VRCamera & the VRSmoothCamera is to be expected when toggling this setting, but this odd correction/lerp of the previous m_Rotation value should not be present.

mtschoen-unity:
Fixed by #442

PR #442, opened 2017-11-30
Fix Presentation/SmoothCamera enabled rotation

dunity:
### Purpose of this PR

Prevent Presentation/SmoothCamera from performing a very fast (shortest path) rotation correction when being enabled : https://github.com/Unity-Technologies/EditorVR/issues/441

### Testing status

Tested numerous times; no issues arose in the tests.

### Technical risk

Low: There's not much to this but setting m_Rotation in OnEnable().

### Notes

When toggling the presentation camera setting, the noticeable roll rotation difference/correction between the VRCamera & the VRSmoothCamera is to be expected.  Though, the (now fixed) odd correction/lerp of the previous m_Rotation value should not be present.

### DIFF
--- a/Scripts/Helpers/VRSmoothCamera.cs
+++ b/Scripts/Helpers/VRSmoothCamera.cs
@@ -79,6 +79,13 @@ namespace UnityEditor.Experimental.EditorVR.Helpers
             m_Rotation = transform.localRotation;
         }
 
+        void OnEnable()
+        {
+            // Prevent camera from performing a very fast (shortest path) rotational correction when being enabled
+            if (m_VRCamera)
+                m_Rotation = m_VRCamera.transform.localRotation;
+        }
+
         void LateUpdate()
         {
             m_SmoothCamera.CopyFrom(m_VRCamera); // This copies the transform as well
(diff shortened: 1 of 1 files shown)


Scripts/Helpers/VRSmoothCamera.cs
```diff
@@ -79,6 +79,13 @@ void Start()
             m_Rotation = transform.localRotation;
         }
 
+        void OnEnable()
+        {
+            // Prevent camera from performing a very fast (shortest path) rotational correction when being enabled
+            if (m_VRCamera)
+                m_Rotation = m_VRCamera.transform.localRotation;
```
mtschoen-unity: Why not Start? And when would VRCamera be null?

dunity: @mtschoen-unity  It's in OnEnable because the VRCam's current localRotation needs to be fetched and set as the new starting position to (smooth)lerp against, each time the presentation cam is enabled.  If this isn't performed, the camera rotation will perform quick "jumps" from the previously cached local rotation value.

dunity: m_VRCamera is null in this case when starting EXR

mtschoen-unity:

PR #443, opened 2017-12-02
Fix exception when loading Poly workspace in project without Poly Toolkit

mtschoen-unity:
If you open EditorXR with a serialized Poly workspace in a project without the PolyToolkit, you get an exception trying to load the previous state. The issue is that the PolyWorkspace stub is just a plain monobehaviour and we can't cast it to IWorkspace.


mtschoen-unity:
Manual merge successful. Performing GitHub merge.

PR #444, opened 2017-12-05
Undo patch fix

mtschoen-unity:
Fixes the frequent "Driving RectTransform" undo actions registered by RectTransform by setting a static bool in a patched version of UnityEngine.CoreModule.dll.

To apply the patch, manually unzip the patched dll and overwrite the version in
<Unity>\Editor\Data\Managed\UnityEngine

We will need to provide patched versions for any Unity version we support. Thankfully the source modification basically just adds the static bool and checks it in two places (and adds the empty class for the CCU to pick up on). We will probably only support major releases until we find a more permanent solution. And we'll want to add patch instructions to the docs.

For the best testing experience, merge this into features/jono/undo-redo-flick (and copy the dll into Program Files, of course).

### DIFF
--- a/Scripts/Core/EditorVR.cs
+++ b/Scripts/Core/EditorVR.cs
@@ -12,6 +12,7 @@ using UnityEngine;
 using UnityEngine.InputNew;
 
 [assembly: OptionalDependency("PolyToolkit.PolyApi", "INCLUDE_POLY_TOOLKIT")]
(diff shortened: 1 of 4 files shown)


mtschoen-unity:
Huh. Lucky it works in p2 even! The problem with this being UnityEnigne.Core is that it's much more likely to conflict with native changes. I'm pretty sure we need a new dll for .3 still

mtschoen-unity:
This has been merged into features/jono/undo-redo-flick

PR #445, opened 2017-12-05
Fix Miniworld Snapping bug

### LINKED ISSUE #268: Objects interpolate position when entering the miniworld breaks snapping
If you move an object into the miniworld while it is snapping, it will "fly into place" instead of popping

mtschoen-unity:
**Purpose of this PR**
Fixes #268. Because objects interpolate position when they are snapping, we had an undesired behavior where objects entering the minworld, and immediately engaging snapping, would quickly fly into place in a jarring fashion. 

The fix is to hard-set the snap position the first frame they enter the miniworld, which maintains the "this is the same object" illusion.

**Testing status**

Tested on Vive in the Polygon Adventure scene. To reproduce the issue pre-fix, just grab an object, find a place in the mini world where it will snap right at the edge of the miniworld volume, and move the object in and out of the miniworld. You should see it fly away from your hand and back into view within the miniworld. With the fix applied, it should just "pop" into the miniworld in the snapped position

**Technical risk**

Low: This only affects the frame when an object enters the miniworld

**Comments to reviewers**

The change to TransferHeldObjects is not represented in this repro case, but the same hard-set makes sense in that situation.

### DIFF
--- a/Scripts/Core/EditorVR.MiniWorlds.cs
+++ b/Scripts/Core/EditorVR.MiniWorlds.cs
@@ -383,8 +383,6 @@ namespace UnityEditor.Experimental.EditorVR.Core
                             directSelection.ResumeGrabbers(incomingPreview.node);
                         }
 
-                        miniWorldRay.UpdatePreview(); // Otherwise the object is in the wrong position for a frame
-
                         if (!isContained)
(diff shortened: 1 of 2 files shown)


mtschoen-unity:

PR #446, opened 2017-12-06
Serialized Feedback Editor

mtschoen-unity:
**Purpose of this PR**
There was a need to clear out the serialized data that stores the number of visible feedback presentations while testing. For example, if I want to show the system to a new user, I might want them to have a "fresh start," or I might want to just enable feedback for one or two features.

This gives us a way of viewing and editing the serialized data for feedback.

**Testing status**

Tested on Vive and Rift in various scenes.  Tested starting EXR without activating the controllers, and with activating controllers but not engaging feedback. Found and fixed a bug where feedback data could be cleared, because we were always returning a SerializedFeedback, even if the nodes themselves were null.

**Technical risk**

Low: Only changes to actual EXR stuff were fixes/refactors. Adds an editor window.

**Comments to reviewers**

I had to expose a bunch of stuff as internal to get static access or access without EXR running. There is a little bit of code duplication as a result, and because Feedback Module is designed around being instantiated and working with EXR systems.


Scripts/Proxies/Data/SerializedProxyFeedback.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Contains per-request persistent data
+    /// </summary>
+    [Serializable]
+    class RequestData
+    {
+        readonly Action m_OnBecameVisible;
```
dunity: Move below properties

Editor/ProxyFeedbackEditor.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Collections.Generic;
+using UnityEditor.Experimental.EditorVR.Modules;
+using UnityEditor.Experimental.EditorVR.Proxies;
+using UnityEngine;
+
+namespace UnityEditor.Experimental.EditorVR.UI
+{
+    sealed class ProxyFeedbackEditor : EditorWindow
+    {
+        Dictionary<Type, SerializedProxyFeedback> m_SerializedFeedback = new Dictionary<Type, SerializedProxyFeedback>();
```
dunity: make readonly

Editor/ProxyFeedbackEditor.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        static void ClearData()
+        {
+            var preferences = SerializedPreferencesModule.DeserializePreferences(Core.EditorVR.serializedPreferences);
+            if (preferences == null)
+                return;
+
+            foreach (var kvp in new Dictionary<Type, SerializedPreferencesModule.SerializedPreferenceItem>(preferences.items))
```
stella3d: is it necessary to create a new dict ?  `preferences.items` is already a `Dictionary<Type, SerializedPreferenceItem>`

mtschoen-unity: We are removing elements

mtschoen-unity:
Oh yeah I forgot to ask/mention: is anyone else seeing errors all over ProxyFeedbackEditor within Visual Studio? It seems like it's ignoring the "InternalsVisibleTo" that we set up to access Engine stuff from an Editor class. If everyone else is getting it, I'll report a bug

Scripts/Proxies/Data/SerializedProxyFeedback.cs
mtschoen-unity: This is addressed. Not sure why it's not hidden.

mtschoen-unity:
Manual merge successful. Performing Github merge

PR #448, opened 2017-12-08
Misc Text Mesh Pro fixes

mtschoen-unity:
The Settings Menu Items still used regular Text.
Also, the Snapping Settings Menu was using GetComponent<Text> to gray out button texts--switched it to TextMeshProUGUI

I didn't address the font material issue because it sounds like there's still a bit of debate around that.

### DIFF
--- a/Scripts/Modules/FeedbackModule/SettingsUI.prefab
+++ b/Scripts/Modules/FeedbackModule/SettingsUI.prefab
@@ -20,7 +20,7 @@ GameObject:
   m_Component:
   - component: {fileID: 224171904768544026}
   - component: {fileID: 222812621535630740}
-  - component: {fileID: 114103337935952270}
+  - component: {fileID: 114454282082524566}
   m_Layer: 5
   m_Name: Label
   m_TagString: Untagged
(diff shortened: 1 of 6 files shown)


Scripts/Modules/FeedbackModule/SettingsUI.prefab
```diff
@@ -320,6 +399,6 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0, y: 1}
   m_AnchorMax: {x: 0, y: 1}
-  m_AnchoredPosition: {x: 0, y: 411.55017}
-  m_SizeDelta: {x: 263.2, y: 45}
+  m_AnchoredPosition: {x: 0, y: 0}
```
stella3d: is the size delta / position change on purpose ?

mtschoen-unity: hm. No it isn't, but it's positioned by a grid layout, so it shouldn't matter. I was resetting their positions before applying the prefabs, so we must just not have done this the first time saving this prefab.

mtschoen-unity:
Manual merge successful. Performing github merge

PR #456, opened 2018-01-02
Drag-to-assign for Assets

stella3d:
Support assigning a variety of assets to their relevant components on objects in the scene.  
Should enable a few more things to be done in EditorXR, particularly because of being able to add any `MonoBehaviour`.

All of these assets are assigned by dragging them out of the project view, direct selecting an object with the hand holding the preview, and then releasing the trigger.

Asset assignment types this supports:

1) Add any `MonoBehaviour` to an object.
2) Assign an `AudioClip` to an object.  If no `AudioSource` is present on the object, one is added
3) Assign a `Material` to an object's `Renderer`.
4) Assign a `Shader` to an instance of the 1st `Material` on an object's `Renderer`
5) Assign a `PhysicMaterial` to an object's `Collider`.
6) Assign a `Font` to a `TextMesh` on an object.


Workspaces/ProjectWorkspace/Scripts/AssetGridItem.cs
```diff
@@ -15,7 +15,7 @@
 namespace UnityEditor.Experimental.EditorVR.Workspaces
 {
     sealed class AssetGridItem : DraggableListItem<AssetData, string>, IPlaceSceneObject, IUsesSpatialHash,
-        IUsesViewerBody, IRayVisibilitySettings, IRequestFeedback, IRayToNode
+        IUsesViewerBody, IRayVisibilitySettings, IRequestFeedback, IRayToNode, IUsesDirectSelection, IGetPreviewOrigin
```
mtschoen-unity: Remove redundant IRayToNode and IGetPreviewOrigin (part of DraggableListItem)

Workspaces/ProjectWorkspace/Scripts/AssetGridItem.cs
```diff
@@ -361,6 +361,24 @@ protected override void OnDragEnded(BaseHandle handle, HandleEventData eventData
...
+                            break;
+                        case "Material":
+                            PlaceMaterial(rayOrigin, data);
+                            break;
+                        case "Script":
+                            PlaceScript(rayOrigin, data);
+                            break;
+                        case "Shader":
+                            PlaceShader(rayOrigin, data);
+                            break;
```
mtschoen-unity: Please also refactor Prefab/Model case into its own method

Workspaces/ProjectWorkspace/Scripts/AssetInstantiation.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System;
+using UnityEngine;
+using UnityEditor.Experimental.EditorVR.Data;
+
+namespace UnityEditor.Experimental.EditorVR.Utilities
```
mtschoen-unity: Move this file to Utilities folder?

Workspaces/ProjectWorkspace/Scripts/AssetInstantiation.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    internal static class AssetInstantiation
```
mtschoen-unity: InstantiationUtils? AssetDropUtils?

Workspaces/ProjectWorkspace/Scripts/AssetInstantiation.cs
```diff
@@ -0,0 +1,109 @@
...
+        const string k_AudioClipAttachUndoLabel = "Add Audio Clip";
+        const string k_AssignFontUndoLabel = "Assign Font";
+        const string k_AttachScriptUndoLabel = "Add Script";
+        const string k_AssignMaterialUndoLabel = "Assign Material";
+        const string k_AssignPhysicMaterialUndoLabel = "Assign Physic Material";
+        const string k_AssignMaterialShaderUndoLabel = "Assign Material Shader";
+
+        internal static AudioSource AttachAudioClip(GameObject go, AssetData data)
+        {
+            Undo.RecordObject(go, k_AudioClipAttachUndoLabel);
```
mtschoen-unity: Are you sure this is the proper way to do the undo here? I figured we would do a RegisterCreatedObject on the new component, otherwise, just setting the field and incrementing the group should do? Might we still need to RegisterCreatedObject separately?

mtschoen-unity: Also, thanks for thinking about undo support 👍

stella3d: everything must Undo :)

Workspaces/ProjectWorkspace/Scripts/AssetInstantiation.cs
stella3d: changed to AssetDropUtils

Workspaces/ProjectWorkspace/Scripts/AssetInstantiation.cs
stella3d: ✅

stella3d:
on the idea of the fallback / catch-all for assigning:  I think it's a good idea, but i'd like to do it another PR.

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+            var renderer = go.GetComponent<Renderer>();
+
+            if (renderer != null)
+            {
+                Undo.RecordObject(go, k_AssignMaterialUndoLabel);
+                // this warns that we're leaking materials into the scene,
+                // and creates a new instance, but we don't want to change
+                // the shader on the shared material here.
+                var shader = (Shader)data.asset;
+                renderer.material.shader = shader;
```
AndrewTHEManeri: Do we want to create a duplicate material and change the shader there before assigning it?  Might as well put the new stuff in an expected place for the user

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                return collider.material;
+            }
+
+            return null;
+        }
+
+        internal static Font AssignFontOnChildren(GameObject go, AssetData data)
+        {
+            var text = go.GetComponentInChildren<TextMesh>();
```
AndrewTHEManeri: What if there is more than one child?

Scripts/Utilities/AssetDropUtils.cs
mtschoen: We discussed this issue. The problem is that creating any kind of asset will cause frame drops. We decided a "good enough" solution is to leak the material into the scene and bypass the error.

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+            Undo.IncrementCurrentGroup();
+            return animation.clip;
+        }
+
+        internal static AudioClip AttachAudioClip(GameObject go, AssetData data)
+        {
+            var source = go.GetComponent<AudioSource>();
+            if (source == null)
+                source = Undo.AddComponent<AudioSource>(go);
+
```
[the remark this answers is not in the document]
mtschoen: I think it records automatically because it's a serialized property. Haven't tested, though.

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+{
+    internal static class AssetDropUtils
+    {
+        const string k_AudioClipAttachUndoLabel = "Add Audio Clip";
+        const string k_AssignAnimationClipUndoLabel = "Add Animation Clip";
+        const string k_AssignVideoClipUndoLabel = "Assign Video Clip";
+        const string k_AssignFontUndoLabel = "Assign Font";
+        const string k_AttachScriptUndoLabel = "Add Script";
+        const string k_AssignMaterialUndoLabel = "Assign Material";
+        const string k_AssignPhysicMaterialUndoLabel = "Assign Physic Material";
```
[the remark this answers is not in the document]
mtschoen: It's called a "physic material" in the create menu

Scripts/Utilities/AssetDropUtils.cs
AndrewTHEManeri: I wonder if we could 'pool' dummy materials in advance and use those - would still have to find a good time to make those however

mtschoen: We'd have to actually write them to disk, and then clean up the unused ones afterward. And you'd have to either make an unreasonable amount, or you might run out.  I think the best course is to just wait for async asset database and make sure we can touch it without locking the UI

mtschoen: Another option is to do the opposite: keep track of any "leaked" materials and write them all out when you close EXR. Since we needed a similar thing for the Poly workspace, it might make sense to set up an "egress" event where systems can do these one-time-on-shutdown actions

Scripts/Utilities/AssetDropUtils.cs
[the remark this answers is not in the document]
mtschoen-unity: It's called a "physic material" in the create menu

(sorry for the spam. Used the wrong account)

Scripts/Utilities/AssetDropUtils.cs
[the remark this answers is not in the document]
stella3d: i need to give the whole Undo treatment a test pass.

Scripts/Utilities/AssetDropUtils.cs
stella3d: i originally had it that way, and for some odd reason, setting a new font on an object that only had one textmesh produced garbled text somehow ?  once i just changed back to a single component, it went away.

I'll revisit that.  that bug seems unlikely and maybe unrelated.

Scripts/Utilities/AssetDropUtils.cs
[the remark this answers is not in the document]
stella3d: yeah - i'm just mimicking what the asset is called, `PhysicMaterial`:
https://docs.unity3d.com/Manual/class-PhysicMaterial.html

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -0,0 +1,132 @@
...
+        const string k_AssignMaterialShaderUndoLabel = "Assign Material Shader";
+
+        internal static AnimationClip AttachAnimationClip(GameObject go, AssetData data)
+        {
+            var animation = go.GetComponent<Animation>();
+            if (animation == null)
+                animation = Undo.AddComponent<Animation>(go);
+
+            Undo.RecordObject(animation, k_AssignAnimationClipUndoLabel);
+            animation.clip = (AnimationClip)data.asset;
```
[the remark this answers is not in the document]
stella3d: oh, i didn't realize you could assign multiple clips to an Animator.  in that case, yes.

stella3d:
@AndrewTHEManeri i have some idea of how we could implement generic 1:1 assignments without too much trouble - if you think it's worth having for this, i'll tackle it before merge.

mtschoen-unity:
Manual merge successful. Performing Github merge.

Note: My final commit (4f1d6f5) cleans up an oversight on #460. I was testing whether it would work to update the inspector on every hierarchy change, but this is not viable as too many other actions (like tooltips) cause hierarchy updates. With this functionality available, I added a call to UpdateInspectors on dropping references to make use of that feature.

PR #457, opened 2018-01-04
Fix bug where closing a workspace while grabbing it locked grip input permenantly

kyleunity:
Fixed the bug by changing RemoveInputProcessor to also remove its controls from m_LockedControls so that input is not kept locked when an input processor is removed (such as closing the workspace).


AndrewTHEManeri:

PR #459, opened 2018-01-05
Bring back Vive "pip"

mtschoen-unity:
**Purpose of this PR**
When we switched to a SerializedObject for ProxyAffordanceMaps, we lost the ability to animate or control visibility on two transforms for a given control ID. The issue is mapping the fields on the prefab to the definitions. The entire map is a single reference, and for the most part things match up 1:1 for each control ID.

The solve is just to use arrays instead of single references for the transforms, renderers, and definitions. We match element 0 to element 0 and so on.  @AndrewTHEManeri and I discussed a future improvement involving custom editors to essentially auto-fill an array index on the elements themselves so that they would stand up to deletions/re-orders, but I'm not sure that is needed at this point.

**Testing status**

Tested on Vive and Rift. Wiggled the buttons and tested the touchpad.

**Technical risk**

Low: Minor code changes and serialized data changes. There are some warnings which we might want to add to help identify bad setup.

**Comments to reviewers**

The highlight will flicker as you cross over the X/Y axis on Vive, which is annoying but I couldn't find the input for "the user is touching the vive." It would require changes to input which are outside the scope of this PR.

### DIFF
--- a/Prefabs/Proxies/LeftTouch.prefab
+++ b/Prefabs/Proxies/LeftTouch.prefab
@@ -2256,45 +2256,65 @@ MonoBehaviour:
     type: 2}
   m_Affordances:
   - m_Control: 2
-    m_Transform: {fileID: 4023718469717780}
-    m_Renderer: {fileID: 13787692}
-    m_Material: {fileID: 0}
+    m_Transforms:
+    - {fileID: 4023718469717780}
+    m_Renderers:
+    - {fileID: 13787692}
+    m_Materials: []
     m_Tooltips:
     - {fileID: 114015681421713280}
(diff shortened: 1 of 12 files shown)


mtschoen-unity:
That helps keep the Transform/renderer count in sync, although we technically don't need it to be. Animation definitions use transforms, and visibility definitions use renderers. There's no real need for there to be the same number of both.

The issue is _relating_ the transforms to their definitions.

jono-unity:
@mtschoen-unity ah, fair enough; in that case, 👍

mtschoen-unity:
Manual merge successful. Performing GitHub merge.

PR #460, opened 2018-01-05
IUpdateInspectors

mtschoen-unity:
### Purpose of this PR
Add an interface for manually updating inspectors, i.e. when adding components (see #456)

### Testing status
Tested by adding code in Selection tool to call UpdateInspectors on trigger press. Update is called successfully.

### Technical risk
Low: Does not affect existing functionality

### Comments to reviewers
We should look to use internal callbacks wherever possible, but I didn't find one that updates just when the inspector updates (without relying on an actual Inspector window). This will have to do, for now.


mtschoen-unity:
I forgot to mention here that I also made a slight modification to ListViewController, etc. to avoid touching SetSiblingIndex when we don't need to. This is primarily what was contributing to the HierarchyChanged spam. You guys probably noticed, but I just realized I didn't mention it in the description.

mtschoen-unity:

PR #461, opened 2018-01-16
Windows Mixed Reality Support

SimonDarksideJ:
Suggest to create a new Feature branch for dev (development/WindowsMMixedReality)

Initial Check-in of WMR support (non-functional)

* EditorVR seems to have trouble to keep running
* Controllers do not visualise


mtschoen-unity:
Hi there!

Thanks for submitting a PR. Unfortunately, we will not be accepting these changes.  It may not have been possible at the time when this was originally submitted, but it is currently possible to use a Windows MR headset and controllers through SteamVR. The controllers show up as Vive controllers in the EditorXR, which is a little awkward, but all of the button mappings still work. This should serve as an interim solution until we prioritize actual support for this hardware.

PR #464, opened 2018-01-23
Remove filter for VR device (3dRudder)

Mvivant:
I have one question:
* Is it possible to remove the ray on proxy? Because I have created a proxy for the 3dRudder device but the proxy created a ray and for the foot-controller it doesn't make sense. It's just a proxy for device that used a locomotion tool.

### DIFF
--- a/Scripts/Modules/DeviceInputModule.cs
+++ b/Scripts/Modules/DeviceInputModule.cs
@@ -67,7 +67,7 @@ namespace UnityEditor.Experimental.EditorVR.Modules
             for (int i = 0; i < devices.Count; i++)
             {
                 var device = devices[i];
-                if (device is VRInputDevice && device.tagIndex != -1)
+                if (/*device is VRInputDevice && */device.tagIndex != -1)
                     m_SystemDevices.Add(device);
             }
 


amirebrahimi:
One idea is that you could add a default tool for the device that makes use of IRayVisibilitySettings to turn off the ray.

Mvivant:
@amirebrahimi Thanks! it works 👍 

I don't know how get access for user to my speed settings (Vector3) in VR ? in a menu or something like that? Is it possible ?

amirebrahimi:
You may be able to use https://github.com/Unity-Technologies/EditorVR/blob/development/Scripts/Interfaces/ISettingsMenuItemProvider.cs. Take a look at other tools that make use of it.

Mvivant:
Thanks but I have the last version of XEditor 0.1.1 and Unity 2017.2.0p2 and the text doesn't appear on settings menu I have a missing monoScript on all Text, I delete all package and re-import but the issue is still here

Mvivant:
Hi, 
How can add Reviewers or Assignees ?

AndrewTHEManeri:
I'm not sure you have permission to do that, I've gone ahead and done it

Mvivant:
Hi,

Someone can take a look on this PR ?
Thanks!

PR #467, opened 2018-02-19
2018 compile fixes

stella3d:
adds a few editor defines , one i missed during the last PR and one for ComponentUtils.

also fixes the BuildPlayer test to work in Unity 2018, using the new BuildReport object instead of a string.

### DIFF
--- a/Scripts/Utilities/AssetDropUtils.cs
+++ b/Scripts/Utilities/AssetDropUtils.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#if UNITY_EDITOR
+using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Video;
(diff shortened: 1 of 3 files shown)


mtschoen-unity:
So far, the thinking has been that we just move with the Unity release cycle, and "don't look back" to supporting older versions.  This would mean that we hold off on these changes until 2018.1 comes out of beta.

These changes seem to be pretty minimal, which wasn't the case moving from 5.6 -> 2017.2. I'm down to support 2017.2 going forward until it becomes impractical to do so. @AndrewTHEManeri, do you have thoughts on this?

Also, if this becomes the new policy (support all versions >= 2017.2) we should update the Readme to say that EXR requires 2017.2+

Scripts/Utilities/AssetDropUtils.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#if UNITY_EDITOR
```
mtschoen-unity: Probably not in this PR, but we should switch to asmdef files for this purpose

stella3d: i agree completely.  and yeah, we should do it in a PR that does that and only that for the whole project.

stella3d:
my thinking is that the officially supported version is of course still 2017.x, but these are some very minimal changes that make it work in 2018.x for our users who use the beta (like me).  just defines and a test fix.

PR #471, opened 2018-04-13
Expand support for copying all enabled image effects on the main camera

dunity:
**Purpose of this PR**
This improvement adds support for copying all _enabled_ images effects on the _Main_ Camera, to EditorXR's VRCamera.  Previously, only the V1 post processing stack was supported.

**Testing status**
Tested using the V1 post processing stack, the V2 post processing stack, and a small assortment of Asset Store image effects.  Tested with both single, and multi-effect copying.

**Technical risk**
Low - This more generalized approach to detecting & copying potential image effects has revealed an issue that occurs when copying certain image effects.  One of the effects I tested against caused Unity to crash every time (Amplify Occlusion).  Though, all others copied as intended, and performed well (aside from the viewer scale issue I describe below).

**Notes**
When performing viewer scaling, there can be artifacts visually present in certain image effects.  Such effects should be re-written to support realtime viewer scaling.  For example, ambient occlusion in the V1 stack presents some visual artifacts when scaling, whereas V2 doesn't.  We should consider adding a rotation-only toggle to viewer scale, in order to alleviate the issue for the image effects that will not add such support, or due to the nature of the effect, cannot add such support without a performance penalty.

**_EDIT_**
To handle for potential image effect copy crashes, or a case in which you still want the camera settings copied, but not the enabled image effects, I added a new corresponding inspector bool/checkbox to EditorXR editing contexts. If set to TRUE on an EditorXR context, enabled image effects on the Main Camera will be copied.

### DIFF
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -210,18 +210,6 @@ namespace UnityEditor.Experimental.EditorVR.Core
                                 targetMethodFound = componentBaseType.GetMethod(targetMethodNames[i], bindingFlags) != null;
                         }
 
-                        // Check nested types for target methods
-                        if (!targetMethodFound)
-                        {
-                            var nestedTypes = componentInstanceType.GetNestedTypes();
-                            foreach (var nestedType in nestedTypes)
-                            {
-                                targetMethodFound = nestedType.GetMethod(targetMethodNames[i], bindingFlags) != null;
-                                if (targetMethodFound)
-                                    break;
-                            }
-                        }
-
                         if (targetMethodFound)
                             break;
                     }
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -145,7 +145,8 @@ namespace UnityEditor.Experimental.EditorVR.Core
 
             s_ExistingSceneMainCamera = Camera.main;
             // TODO: Copy camera settings when changing contexts
-            if (EditingContextManager.defaultContext.copyExistingCameraSettings && s_ExistingSceneMainCamera && s_ExistingSceneMainCamera.enabled)
+            var defaultContext = EditingContextManager.defaultContext;
+            if (defaultContext.copyMainCameraSettings && s_ExistingSceneMainCamera && s_ExistingSceneMainCamera.enabled)
             {
                 GameObject cameraGO = EditorUtility.CreateGameObjectWithHideFlags(k_CameraName, HideFlags.HideAndDontSave);
                 m_Camera = ObjectUtils.CopyComponent(s_ExistingSceneMainCamera, cameraGO);
@@ -184,7 +185,7 @@ namespace UnityEditor.Experimental.EditorVR.Core
             m_CameraRig.position = headCenteredOrigin;
             m_CameraRig.rotation = Quaternion.identity;
 
-            if (s_ExistingSceneMainCamera)
+            if (s_ExistingSceneMainCamera && defaultContext.copyMainCameraImageEffects)
             {
                 var cameraGameObject = m_Camera.gameObject;
                 var potentialImageEffects = s_ExistingSceneMainCamera.GetComponents<MonoBehaviour>();
@@ -197,8 +198,6 @@ namespace UnityEditor.Experimental.EditorVR.Core
                     var targetMethodFound = false;
                     for (int i = 0; i < targetMethodNames.Length; ++i)
                     {
-                        // Each of the three checks is performed to catch the various image effect variants I've tested against
-                        // Each check catches a different case that was encountered during testing
                         // Check isntanced type for target methods
                         targetMethodFound = componentInstanceType.GetMethod(targetMethodNames[i], bindingFlags) != null;
 
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -216,9 +216,6 @@ namespace UnityEditor.Experimental.EditorVR.Core
                             var nestedTypes = componentInstanceType.GetNestedTypes();
                             foreach (var nestedType in nestedTypes)
                             {
-                                if (nestedType == null)
-                                    continue;
-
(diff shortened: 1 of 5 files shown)


Scripts/Core/VRView.cs
```diff
@@ -189,6 +184,70 @@ public void OnEnable()
...
+                    if (targetMethodFound)
+                    {
+                        try
+                        {
+                            // During testing, some image effects caused Unity to crash when copied
+                            ObjectUtils.CopyComponent(potentialImageEffect, cameraGameObject);
+                        }
+                        catch (Exception e)
+                        {
+                            Console.WriteLine(e);
```
mtschoen-unity: Why not Debug.Log? Also, what exceptions? What Unity version? I'm curious about why catching them here would prevent a crash.

Scripts/Core/VRView.cs
```diff
@@ -189,6 +184,70 @@ public void OnEnable()
...
+                            foreach (var nestedType in nestedTypes)
+                            {
+                                if (nestedType == null)
+                                    continue;
+
+                                targetMethodFound = nestedType.GetMethod(targetMethodNames[i], bindingFlags) != null;
+                                if (targetMethodFound)
+                                    break;
+                            }
+                        }
```
mtschoen-unity: Will the nested types actually get their methods called? What image effect is this necessary for?

Scripts/Core/VRView.cs
```diff
@@ -189,6 +184,70 @@ public void OnEnable()
...
+                        // Check isntanced type for target methods
+                        targetMethodFound = componentInstanceType.GetMethod(targetMethodNames[i], bindingFlags) != null;
+
+                        // Check base type for target methods
+                        if (!targetMethodFound)
+                        {
+                            var componentBaseType = componentInstanceType.BaseType;
+                            if (componentBaseType != null)
+                                targetMethodFound = componentBaseType.GetMethod(targetMethodNames[i], bindingFlags) != null;
+                        }
```
mtschoen-unity: Add a Utils method to recursively check base types

Scripts/Core/VRView.cs
```diff
@@ -189,6 +184,70 @@ public void OnEnable()
             m_CameraRig.position = headCenteredOrigin;
             m_CameraRig.rotation = Quaternion.identity;
 
+            if (s_ExistingSceneMainCamera)
+            {
+                var cameraGameObject = m_Camera.gameObject;
+                var potentialImageEffects = s_ExistingSceneMainCamera.GetComponents<MonoBehaviour>();
+                var enabledPotentialImageEffects = potentialImageEffects.Where(x => x != null && x.enabled);
+                var targetMethodNames = new [] {"OnRenderImage", "OnPreRender", "OnPostRender"};
+                var bindingFlags = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static;
```
mtschoen-unity: Not sure static is necessary

Scripts/Core/VRView.cs
```diff
@@ -189,6 +184,70 @@ public void OnEnable()
...
+                        }
+
+                        // Check nested types for target methods
+                        if (!targetMethodFound)
+                        {
+                            var nestedTypes = componentInstanceType.GetNestedTypes();
+                            foreach (var nestedType in nestedTypes)
+                            {
+                                if (nestedType == null)
+                                    continue;
```
mtschoen-unity: Do we really get nulls?

Scripts/Core/VRView.cs
dunity: I'll remove this. After additional testing, a small number of image effects still crash the editor regardless of the try/catch.

Scripts/Core/VRView.cs
dunity: We do not.  I'll remove this.

Scripts/Core/VRView.cs
dunity: I can't find the exact effect that this handled for.  I've since deleted the many effects that I tested against.  Though, after removing this nested type check, the new roster of effects I'm testing with all work as expected (except for Amplify Occlusion, which still crashes).  I'll commit the removal of this block.

Scripts/Core/VRView.cs
dunity: Added to ComponentUtils as "MethodFoundInBaseType"

dunity:
To handle for potential image effect copy crashes, or a case in which you still want the camera settings copied, but not the enabled image effects, I added a new corresponding inspector bool/checkbox to EditorXR editing contexts.  If set to TRUE on an EditorXR context, enabled image effects on the Main Camera will be copied.  <--- @AndrewTHEManeri  @mtschoen-unity

dunity:
Aside from that PR comments have been addressed

dunity:
@mtschoen-unity As for a scene that tests this:  We could, but we'd need to include in the project either a dummy image effect for testing, or add existing image effects to the repo, submod in the post stack, etc.  Either way, the only effect I've encountered that crashes during this process, is Amplify Occlusion.  The other image effects copied & performed as expected.  Lemme know your thoughts/feedback.

mtschoen-unity:
Good call on splitting out camera copy/effects copy.

A slight nitpick: I'd like to see the utility method look more like this:
```
public static MethodInfo GetMethodRecursively(this Type type, string name, BindingFlags bindingAttr)
{
    var method = type.GetMethod(name, bindingAttr);
    if (method != null)
        return method;

    var baseType = type.BaseType;
    if (baseType != null)
        method = type.BaseType.GetMethodRecursively(name, bindingAttr);

    return method;
}
```

As for the test--anything that fails if this breaks is the point. If we need to include the Post stack, we can still compile-out references with the CCU. If we can't do a unit test programmatically, maybe we just need a private repo or something that we can use as a test going forward. I wouldn't know what should/shouldn't work when updating to a new version of Unity, for example. This can go into an issue/future PR.

AndrewTHEManeri:
Manual merge successful :)

PR #472, opened 2018-04-17
Add selective support for copying enabled image effects to the presentation camera

dunity:
**Purpose of this PR**
Add support for copying Main Camera image effects to the presentation camera (separate from the HMD).  A new inspector bool was added to EXR contexts, in order to selectively enable/disable this functionality for either/both the HMD & presentation camera.

The below fields/properties have been added to EditorXR contexts:
_copyMainCameraImageEffectsToHMD_
_copyMainCameraImageEffectsToPresentationCamera_

**Testing status**
Tested using many different post processing image effects.  Tested with both single, and multi-effect copying.

**Technical risk**
Low - Aside from "Amplify Occlusion", I found no issues during testing.  All effects copied to their intended camera, and performed as expected (aside from the viewer scale issue I describe below).

**Notes**
When performing viewer scaling, there can be artifacts visually present in certain image effects. Such effects should be re-written to support realtime viewer scaling. For example, ambient occlusion in the V1 stack presents some visual artifacts when scaling, whereas V2 doesn't. We should consider adding a rotation-only toggle to viewer scale, in order to alleviate the issue for the image effects that will not add such support, or due to the nature of the effect, cannot add such support without a performance penalty.

### DIFF
--- a/Scripts/Core/Contexts/EditorVR.asset
+++ b/Scripts/Core/Contexts/EditorVR.asset
@@ -13,7 +13,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   m_RenderScale: 1
   m_CopyMainCameraSettings: 1
-  m_CopyMainCameraImageEffects: 1
+  m_CopyMainCameraImageEffectsToHmd: 1
+  m_CopyMainCameraImageEffectsToPresentationCamera: 1
   m_DefaultToolStack:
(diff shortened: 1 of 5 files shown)


dunity:
Although duplicate copying of image effects to the presentation camera on enable/disable wasn't occurring; only perform such an image effect copy when the active view's custom preview camera is null (when it is first set).

AndrewTHEManeri:
Manual merge successful, oh yeah

PR #480, opened 2018-06-29
Added footindicator to give users a life-size reference of their own …

yinghuayang:
### Purpose of this PR

A foot indicator is added to give users a reference of their current height. [Gif demo](https://drive.google.com/file/d/1_NsGmldNrfSiq7NZqIGGxYC6YIpYmcRV/view?usp=sharing)

### Testing status

Tested in a scene with locomotion and scaling.

### Technical risk

Low. There is a new prefab in the scene. 

### Comments to reviewers

Temporarily used chicken legs from Poly.


Scripts/Helpers/PlayerFloor.cs
```diff
@@ -0,0 +1,35 @@
+﻿#if UNITY_EDITOR
+using UnityEngine;
```
dunity: Sort usings: (...Core, then ...Extensions, then...Utilities, then UnityEngine)

Scripts/Helpers/PlayerFloor.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        private void Awake()
+        {
+            m_Camera = CameraUtils.GetMainCamera().transform;
+            m_CameraRig = CameraUtils.GetCameraRig();
+        }
+
+        void Update()
+        {
+            floorPosition.x = m_Camera.position.x + 0.05f * this.GetViewerScale();
```
dunity: cache viewer scale value initially, before utilizing it 3X below

Scripts/Extensions/TransformExtensions.cs
```diff
@@ -22,6 +22,24 @@ public static Bounds TransformBounds(this Transform transform, Bounds localBound
...
+        public static Vector3 XZForward(this Transform target)
+        {
+            var forward = target.forward;
+            if ((forward.y * forward.y) >= 0.5f)
+            {
+                forward = -target.up * Mathf.Sign(forward.y);
+            }
+            else
+            {
+                if (target.up.y < 0.0f)
```
dunity: Convert to "else if" block, instead of an if block inside the else block:
...(else if (target.up.y < 0.0f)
{
    forward = -forward;
})

Scripts/Extensions/TransformExtensions.cs
```diff
@@ -26,20 +26,22 @@ public static Bounds TransformBounds(this Transform transform, Bounds localBound
         public static Vector3 XZForward(this Transform target)
         {
             var forward = target.forward;
-            if ((forward.y * forward.y) >= 0.5f)
+            var up = forward.y;
+            const float mid = 0.5f;
```
dunity: rename to "kMid"

Scripts/Extensions/TransformExtensions.cs
```diff
@@ -26,20 +26,22 @@ public static Bounds TransformBounds(this Transform transform, Bounds localBound
...
+            const float min = 0.0f;
```
dunity: rename to "kMin"

Scripts/Extensions/TransformExtensions.cs
```diff
@@ -26,20 +26,22 @@ public static Bounds TransformBounds(this Transform transform, Bounds localBound
...
             {
-                if (target.up.y < 0.0f)
-                {
-                    forward = -forward;
-                }
+                forward = -forward;
             }
-            forward.y = 0.0f;
+
+            up = 0.0f;
```
dunity: Not sure if this should have changed from forward.y.  Setting up to any value here does nothing.

Scripts/Helpers/PlayerFloor.cs
yinghuayang: I got an error saying that the contextual keyword 'var' may only appear within a local variable when I wrote "var currentScale = this.GetViewerScale()". Should I just declare GetViewScale() instead?

mtschoen-unity: Maybe you were using "var" inline with the assignment? It should work if you do
```var currentScale = this.GetViewerScale();
floorposition.x = ... * currentScale```

AndrewTHEManeri:
Use TextMeshPro text for the floor text instead of TextMesh - will give better quality results

AndrewTHEManeri:
I want to see how the gaze divergence works w/ the floor plane.  If I scale myself up in the adventure scene and put it at torso-height, I tend to start getting double vision because of wanting to focus on the close geo and the distant outlines.  One additional solution might be to tie the visibility of the outlines to the controller shake we have for making the controllers fade in.

PR #481, opened 2018-06-29
Annotation Tool Pressure Sensitiveity + Menu Rework

AndrewTHEManeri:
### Purpose of this PR
General improvements to the annotation tool:
Adds pressure sensitivity, pressure smoothing, and consolidates the various off-hand menus to one menu in the same style as the create primitive tool.  This has a side-effect of simplifying the annotationtool script and prefabs.

Undo/redo menus are no longer suppressed by custom menus, to facilitate that feature being available while the annotation or create primitive menu is up.

### Testing status

Tested in headset in empty and busy scene.  Tested as various scales.  Used every menu option, spatially selected different tools, and cycled through tool options as well.

### Technical risk

Medium - not blocking the undo/redo menu when a custom menu is up could have unexpected consequences.


dunity:
The text on the UI panel is visible from the back of the panel.  This is likely a render-queue issue, and should be remedied.

dunity:
In addition, I've found a reproducible bug in this branch:
- Add the annotation tool to a given hand (via selecting it from the main menu on the other hand)
- On the hand in which the tool was added, the (3-line hamburger)main-menu icon will stay-hidden, and will be replaced instead with a duplicate "AT"(annotation-tool) (TMP)text-string (in addition to the actual "AT" button to the side of it).  The corresponding main-menu functionality will also be inaccessible.

mtschoen-unity:
The Double AT bug and the issue where the CreatePrimitive tool is active when closing the AnnotationTool is present on `development` and I have created an issue for this task

PR #483, opened 2018-07-02
Use assembly definition files; Get play-mode working

AndrewTHEManeri:
### Purpose of this PR

Enables EditorVR support in play-mode, applies assembly definition files for all submodules.  

### Testing status

That's what we're doing here.  Let's be thorough.

### Technical risk

High - the 'editor only' assumption of EXR is going away and we need to be very vigilant looking for unintended side-effects.

### Comments to reviewers

Let's see what kind of velocity we have with all-eyes on everything.  If we're not making much progress we might do a divide-and-conquer approach instead.


amirebrahimi:
I'm happy to help with this, too. If there are a lot of changes, then one suggestion is to merge this into a local dev-branch here on this repo and then perform the PR from there. Otherwise, we'll need to make changes in the original branch.

Scripts/Utilities/ComponentUtils.cs
```diff
@@ -31,12 +31,14 @@ public static class ComponentUtils
         public static T GetOrAddIf<T>(GameObject go, bool option) where T : Component
         {
             T component = go.GetComponent<T>();
+#if UNITY_EDITOR
             if (option && component == null)
                 component = Undo.AddComponent<T>(go);
-
+#endif
```
AndrewTHEManeri: Add the component without undo if in play mode?

Shaders/EVRMeshChain.shader
```diff
@@ -53,7 +53,9 @@
 		// prevent overlapping lines from adding too much color,
 		// we set the alpha value to one after visiting a pixel.
 		Blend OneMinusDstAlpha DstAlpha, One One
-		BlendOp Add, Max
+		// FIXME: There's a bug where the _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A keyword gets set incorrectly causing the alpha to be incorrect and consequently the
```
AndrewTHEManeri: Note to self: Look into this

Scripts/Utilities/MaterialUtils.cs
```diff
@@ -20,8 +20,12 @@ static class MaterialUtils
         /// <returns>Cloned material</returns>
         public static UnityMaterial GetMaterialClone(Renderer renderer)
         {
+#if UNITY_EDITOR
             // The following is equivalent to renderer.material, but gets rid of the error messages in edit mode
             return renderer.material = UnityObject.Instantiate(renderer.sharedMaterial);
+#else
+            return renderer.material;
```
AndrewTHEManeri: Any reason that instantiating the sharedMaterial as above is not appropriate?

AndrewTHEManeri:
@amirebrahimi I'm getting errors about missing the SteamVR assembly and OVR Assembly.  Did you make those manually or are they in updated libraries somewhere?

PR #486, opened 2018-07-03
Use assembly definition files; Get play-mode working

AndrewTHEManeri:
### Purpose of this PR

Enables EditorVR support in play-mode, applies assembly definition files for all submodules.  

### Testing status

That's what we're doing here.  Let's be thorough.

### Technical risk

High - the 'editor only' assumption of EXR is going away and we need to be very vigilant looking for unintended side-effects.

### Comments to reviewers

Let's see what kind of velocity we have with all-eyes on everything.  If we're not making much progress we might do a divide-and-conquer approach instead.


Actions/Copy.cs
```diff
@@ -1,14 +1,16 @@
-﻿#if UNITY_EDITOR
+﻿
 namespace UnityEditor.Experimental.EditorVR.Actions
 {
     [ActionMenuItem("Copy", ActionMenuItemAttribute.DefaultActionSectionName, 5)]
     sealed class Copy : BaseAction
     {
         public override void ExecuteAction()
         {
+#if UNITY_EDITOR
```
AndrewTHEManeri: The cut/copy/paste/duplicate commands we can probably simulate - instantiate a deactivated object on copy (in the clip board), instantiate activate duplicates on paste.

Actions/Undo.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if UNITY_EDITOR
+﻿
 namespace UnityEditor.Experimental.EditorVR.Actions
 {
     [ActionMenuItem("Undo", ActionMenuItemAttribute.DefaultActionSectionName, 2)]
```
AndrewTHEManeri: This is kind of out there thinking - we're already doing an undo-patch for EXR. Any way we can just shoehorn undo/redo in through that patch?

Menus/UndoMenu/UndoMenu.cs
```diff
@@ -120,14 +120,18 @@ public void ProcessInput(ActionMapInput input, ConsumeControlDelegate consumeCon
             var undoRedoPerformed = false;
             if (navigateX < -k_UndoRedoThreshold && (m_TrackpadController || m_PrevNavigateX > -k_UndoRedoThreshold))
             {
+#if UNITY_EDITOR
```
AndrewTHEManeri: Probably should do a utils version of undo/redo so we can avoid these ifdefs or ideally switch in playmode functionality

amirebrahimi:
FYI - there's a change in the mix from Harald Nielsen to assembly definition files that may allow optional dependencies. Until then, we'll need to figure out a solution for our partner SDKs (i.e. Vive and Oculus).

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -303,21 +348,52 @@ internal static EditingContextManagerSettings LoadUserSettings()
...
 
         internal static void SaveUserSettings(EditingContextManagerSettings settings)
         {
+#if UNITY_EDITOR
             File.WriteAllText(k_UserSettingsPath, JsonUtility.ToJson(settings, true));
+#endif
+        }
+
+#if UNITY_EDITOR
+        [ContextMenu("Utility")]
```
AndrewTHEManeri: I'm not sure if this should be user-facing or was just a utility function for debugging input?

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -331,6 +407,10 @@ static void InitializeInputManager()
 
             managers = Resources.FindObjectsOfTypeAll<InputManager>();
 
+//            var gameObjects = Resources.FindObjectsOfTypeAll<GameObject>();
```
AndrewTHEManeri: remove commented code

Scripts/Modules/SelectionModule/SelectionModule.cs
```diff
@@ -135,10 +135,12 @@ GameObject GetGroupRoot(GameObject hoveredObject)
             if (m_GroupMap.TryGetValue(hoveredObject, out groupParent))
                 return groupParent;
 
+#if UNITY_EDITOR
```
AndrewTHEManeri: We could use our instantiation helper to insert a dummy component that can act as a group root in runtime mode

Scripts/Runtime/Selection.cs
```diff
@@ -1,14 +1,65 @@
...
+    public static Object activeObject
+    {
+        get {return s_Objects != null && s_Objects.Length > 0 ? s_Objects[0] : null; }
+        set { s_Objects = value ? new[] { value } : null; }
+    }
+
+    public static GameObject activeGameObject
+    {
+        get {return s_Objects != null && s_Objects.Length > 0 ? s_Objects[0] as GameObject : null; }
+        set { s_Objects = value ? new[] { value } : null; }
```
AndrewTHEManeri: Can s_Objects ever hold components? If so, we might want to handle that case and bubble up the owning game object

Scripts/Utilities/ComponentUtils.cs
```diff
@@ -33,9 +33,11 @@ public static class ComponentUtils
         public static T GetOrAddIf<T>(GameObject go, bool option) where T : Component
         {
             T component = go.GetComponent<T>();
+#if UNITY_EDITOR
             if (option && component == null)
                 component = Undo.AddComponent<T>(go);
-
+#endif
```
AndrewTHEManeri: Add the component without undo if in play mode?

Shaders/EVRMeshChain.shader
```diff
@@ -53,7 +53,9 @@
 		// prevent overlapping lines from adding too much color,
 		// we set the alpha value to one after visiting a pixel.
 		Blend OneMinusDstAlpha DstAlpha, One One
-		BlendOp Add, Max
+		// FIXME: There's a bug where the _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A keyword gets set incorrectly causing the alpha to be incorrect and consequently the
+		// line renderer to not render correctly, so for now this is set to Max, so it shows up. It should be Add.
+		BlendOp Max, Max
```
AndrewTHEManeri: Note to self: Look into this

Scripts/Utilities/MaterialUtils.cs
```diff
@@ -20,8 +20,12 @@ static class MaterialUtils
         /// <returns>Cloned material</returns>
         public static UnityMaterial GetMaterialClone(Renderer renderer)
         {
+#if UNITY_EDITOR
             // The following is equivalent to renderer.material, but gets rid of the error messages in edit mode
             return renderer.material = UnityObject.Instantiate(renderer.sharedMaterial);
+#else
+            return renderer.material;
```
AndrewTHEManeri: Any reason that instantiating the sharedMaterial as above is not appropriate?

AndrewTHEManeri:
@amirebrahimi I'm getting errors about missing the SteamVR assembly and OVR Assembly. Did you make those manually or are they in updated libraries somewhere?

Scripts/Utilities/ComponentUtils.cs
amirebrahimi: Ahh...Good find.

Scripts/Runtime/Selection.cs
amirebrahimi: Yeah, this was a quick hack. Seems to work for our needs at the moment though.

Scripts/Core/Contexts/EditingContextManager.cs
amirebrahimi: This can be removed.

amirebrahimi:
Yeah, those assembly definition files were created manually. I can send them to you if you like. Still need to think about how to patch those SDKs.

Scripts/Core/Contexts/EditingContextManager.cs
amirebrahimi: Yeah, this should be deleted

Menus/UndoMenu/UndoMenu.cs
amirebrahimi: The latest hackweek project of a C# undo system might work.

AndrewTHEManeri:
Yeah, send the files over.  Maybe we could do something like a combination of the serialized object references and CCU?  Determine if an assembly is present, and if so, copy the assembly file into that folder

amirebrahimi:
Use `7z.exe x -oAssets partner-sdk-asmdefs.zip` to overlay files over existing directory structure. This assumes that your zip is in the root of your unity project folder and that the OVR and SteamVR SDKs have not been moved from the original installation.

[partner-sdk-asmdefs.zip](https://github.com/Unity-Technologies/EditorXR/files/2160753/partner-sdk-asmdefs.zip)

Scripts/Utilities/MaterialUtils.cs
amirebrahimi: Unity copies on access by default for renderer.material. Or I might have gotten a warning with the the instantiation of the shared material when in play  mode. Not sure.

amirebrahimi:
@AndrewTHEManeri - You're right. I forgot that I also added AssemblyInfo.cs to OVR:
```
[assembly: InternalsVisibleTo("EXR")]
```

Scripts/Core/VRView.cs
```diff
@@ -139,56 +146,63 @@ public static Coroutine StartCoroutine(IEnumerator routine)
             return null;
         }
 
-        public void OnEnable()
+        public static void CreateCameraRig(ref Camera camera, ref Transform cameraRig)
```
mtschoen-unity: This change is causing a compiler error due to a reference to m_Camera below (I think it's from the merge w/ dev). I can get around it by just commenting out the line.

dunity: I'd also consider switching "m_Camera" to "viewerCamera"; the s_ActiveView's static m_Camera ref.

dunity: I just pushed this quick fix. It warrants additional testing though.

amirebrahimi:
> @amirebrahimi I'm wondering what version of OVR you used during your work on play-mode? AFAIK, only older versions of OVR still have a folder hierarchy similar to that which is contained in "the partner-sdk-asmdefs.zip"?

@dunity - When I go to Tools->Oculus->Update plugin it says 1.22.0. The last zip file I have on disk for OVR Utilities is 1.15.0 though. Did they switch their folders around?

Scripts/Core/VRView.cs
amirebrahimi: This should be `camera` as that is passed in to the function, not viewerCamera.

Actions/Copy.cs
mtschoen-unity: Yeah, we used to do it that way. Worked fine

Actions/Undo.cs
mtschoen-unity: No, the undo system is currently implemented in the editor itself, with C# API hooks. As Amir mentioned, the C# undo hack week project might work

amirebrahimi:
A few updates:
1. Got in touch with Oculus and will provide a patch to them for assembly definition files.
2. Optional assembly definitions will likely go into 2019.1, so possibly we will need to use @mtschoen-unity's solution until then.

Scripts/Utilities/MaterialUtils.cs
mtschoen-unity: This change makes sense to me. This method exists to avoid the warning about using renderer.material in edit mode. In play mode, we can use this functionality without issue.

Tools/LocomotionTool/LocomotionTool.cs
```diff
@@ -228,7 +227,7 @@ void OnDestroy()
             this.RemoveRayVisibilitySettings(rayOrigin, this);
             this.ClearFeedbackRequests();
 
-            if (m_ViewerScaleVisuals)
+            if (m_ViewerScaleVisuals && m_ViewerScaleVisuals.gameObject.activeInHierarchy)
```
mtschoen-unity: @amirebrahimi do you remember why this changed?

Scripts/Runtime/Selection.cs
mtschoen-unity: @amirebrahimi, what more needs to be done here? I was able to select and manipulate an object in play mode.

README.md
```diff
@@ -43,7 +43,12 @@ Optionally, you could add a [git hook for post-checkout](https://ttboj.wordpress
...
 
-We're using `#if UNITY_EDITOR` for now in order to keep our code out of your player builds. We will eventually explore a possible overlap between EditorXR and player builds. We recommend you do the same for your tools if you plan to distribute them to others.
+### Assembly Definitions
+In order to support a variety of platform configurations, and to optionally strip its code out of player builds, EditorXR uses assembly definitions. The EditorXR assembly definition must reference both platform SDK packages (SteamVR or Oculus VR) from the Asset Store. These packages do not include assembly definitions in their current forms, so after importing EditorXR, you must add them.
+
+For easy set-up, EditorXR includes a .unitypackage (`Patches/Dependencies_asmdef.unitypackage`) containing these assembly definitions in the locations they are likely to be needed. This package also contains an assembly definition for the PolyToolkit, which is also referenced by EditorXR.
+
+This is not required for Unity versions 2019.1 and above.
```
amirebrahimi: Have you tested this?

Scripts/Runtime/Selection.cs
amirebrahimi: I think this is good enough for now. It's just a shim that can be replaced maybe at some point if runtime support is ever fully well...supported.

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
```diff
@@ -211,9 +210,9 @@ public LayerMask layerMask
         // Local method use only -- created here to reduce garbage collection
         RayEventData m_TempRayEvent;
 
-        protected override void Awake()
+        protected override void OnEnable()
         {
-            base.Awake();
+            base.OnEnable();
```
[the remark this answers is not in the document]
amirebrahimi: I believe this is with how initialization had to change for the script runtime because all of the default script references now already exist on a prefab to be instantiated from and enabled rather than how they are created from the editor.

Tools/LocomotionTool/LocomotionTool.cs
amirebrahimi: No idea at this point.

Menus/UndoMenu/UndoMenu.cs
amirebrahimi: @mtschoen-unity - you may want to look into this, too --^

README.md
mtschoen-unity: Yep! I tested it in one of my projects that was already mostly set up. One of my tasks for today is to do a clean setup from an empty project, both from github and from a package.

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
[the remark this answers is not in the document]
mtschoen-unity: Aha. Ok I will test going back to Awake to see if/what breaks. We may be able to still use awake if we set the default script references script to have a super-low execution order. This was how we allowed for using FI-dependent code in MARS--the script which injects functionality on everything forces itself into a very low execution order.

Tools/LocomotionTool/LocomotionTool.cs
mtschoen-unity: OK. I will try reverting it and see what happens.

Menus/UndoMenu/UndoMenu.cs
mtschoen-unity: Yes. I will definitely be taking a look at the hack week project for undo/redo. I've tasked it out for now--out of scope for this PR.

amirebrahimi:
@mtschoen-unity - Lemme know how those revert changes go and then I'll approve.

Tools/LocomotionTool/LocomotionTool.cs
mtschoen-unity: I couldn't reproduce any issue around the viewer visuals in edit mode, play mode, or in a player build after reverting this change

Scripts/Modules/MultipleRayInputModule/MultipleRayInputModule.cs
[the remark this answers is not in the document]
mtschoen-unity: I tried reverting all of the modules where Awake changed into OnEnable and I found that Awake was still called in Play Mode and Edit Mode. Oddly, I found that EditorVR.Awake doesn't get called in Play Mode, which it seems is why you changed Awake to Initialize, and call it in the beginning of Start. I'll look into this later, but for now, we can safely revert the modules and leave EditorVR as it is.

mtschoen-unity:
I'd say this is good to go at this point. I've been able to use all workspaces and tools in Edit Mode, a subset of workspaces in Play Mode, and the MiniWorld Workspace and tools in a built player. I haven't tested all actions in the player but we can refine individual functionality in smaller PRs. I have made tasks for a few remaining cleanup items.

The asmdef fix works out pretty nicely. EXR won't compile on first import, but all you have to do is immediately import the asmdef package and you're good to go. We can even include the asmdefs in the release package to avoid this issue for users of the package. The issue will also go away in 2019.1, where missing references are simply ignored and compilation can continue. I've updated the readme to explain this.

There is one minor issue where the EXR object (the one w/ the EditingContextManager) in the open scene gets destroyed right before you enter Play Mode. I think there is some code in the EXR lifecycle that destroys the ECM, but I don't want to get rid of it now in case we end up leaking them or something.  I have created a task to investigate and fix this issue. This does not affect Edit Mode.

Some things I encountered and fixed during testing:
- The DefaultScriptReference script had a slight issue where it would check/create the Resources directory after first trying to instantiate the prefab. Also, I should note that I had to manually add the SpatialMenu to the list of types to collect. We might want to add a catch-all interface for miscellaneous scripts that need their references serialized.
- Oculus Utilities on Asset Store now has assembly definitions, so I've updated our references and the contents of the unity package to contain these assembly definitions in their imported paths
- Something about this branch, or the new Oculus Utilites cause the Touch input axes to not zero out completely, which means nothing ever unlocks controls after being consumed. I added a dead zone check to OVRInputToEvents. Oddly, I was not able to reproduce this issue with latest development and the new Oculus Utilities, but couldn't track down a particular change here or in the input prototype that would have lead to this issue.
- Our TextMeshPro warning was coming up on first import, even though the package is included by default, because of the way the CCU does multiple passes. I was able to suppress the warning with a reflection check on the first pass. I also updated the warning to indicate the package manager instead of Asset Store. I checked that the TMP package works back in 2017.2, and it does!
- The Getting Started guide needed updating, and I also added a bit about play mode in the readme.
- I had to change some GUIDs on our copies of the Touch controller assets so that they don't get overwritten by the Oculus Utilities package import. What was happening is that the package importer found the assets in EXR, updated them, and broke texture references. It also modifies files in the repo, which is less than ideal. A number of the files that were being modified weren't even used, so I deleted them. That's why `1b360f2` deletes a bunch of .fbx files.
- My plan of compiling out the Project, etc. workspaces didn't work out for Play Mode. I added an `EditorOnlyWorkspaceAttribute` to mark workspaces that we don't want available in Play Mode. Eventually we can probably let contexts handle this.
- SerializedPreferences weren't used in player builds, but they seem to work anyway. I was hoping to resolve an issue in the AnnotationTool where it would hit null refs if you had never serialized its preferences. I had to just check against that case because of the order in which preferences are loaded.

PR #496, opened 2018-08-29
Display enabled/disabled state in Device View

dunity:
A simple PR that adds support for displaying text in the VRView's device view, denoting that the device view is either enabled or disabled.


Scripts/Core/VRView.cs
```diff
@@ -404,7 +404,9 @@ void OnGUI()
             m_ToggleDeviceViewRect.width = width;
             m_PresentationCameraRect.y = height - m_PresentationCameraRect.height;
 
-            if (GUI.Button(m_ToggleDeviceViewRect, "Toggle Device View", EditorStyles.toolbarButton))
+            const string kToggleDeviceViewEnabled = "Toggle Device View : Enabled";
+            const string kToggleDeviceViewDisabled = "Toggle Device View : Disabled";
+            if (GUI.Button(m_ToggleDeviceViewRect, m_ShowDeviceView ? kToggleDeviceViewEnabled : kToggleDeviceViewDisabled, EditorStyles.toolbarButton))
```
mtschoen-unity: Are there any HIG or general guidelines on toggles like this? Should we switch to a different UI control? I do like the idea of having this indicate the state to make this more discoverable.

PR #498, opened 2018-09-10
Add spatial-scrolling support to the SpatialMenu

dunity:
This PR adds X/Y horizontal/vertical spatial scrolling support to the EditorXR "SpatialMenu". 

This implementation handles +- X/Y horiz/vert axis spatial scrolling, relative to the vector defined by the location of the rayOrigin and the HMD, upon display of the SpatialMenu.

This functionality allows a user the ability to move their hand left/right to cycle though top-level menu items (after initially opening the SpatialMenu).  Trigger can then be used to select a highlighted element.  Then, after selecting a highlighted top-level menu element (actions/workspaces/tools), when in a sub-menu, the user can move their wrist up and down (with their elbow staying in-place), in order to scroll vertically through the sub-menu elements.  Again trigger will allow for element selection.  At any point the user can use a ray to disable spatial-select mode, or use the cyclical rotation input mode to select menu elements.  Grip cancels, and returns to previous menu level, as usual.

Additional PR description details will be added later


dunity:
After having just merged in latest Development, I'll refine this PR a bit before it should be considered for PR review.

mtschoen-unity:
@dunity wants a little more time on this, so I'm closing the PR for now.

PR #500, opened 2018-10-14
Minimal context

amirebrahimi:
This is branched off of the TMP upgrade, so those changes will need to land first.

Anyhow, this PR adds a new Minimal context, which strips out the default main menu, radial menu, and all workspaces expect the project workspace.

Additionally, there is a new feature of auto-opening EXR when the headset is on a person's head. This can be turned off in the Edit->Preferences->EditorXR.


Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -7,6 +7,8 @@
 using UnityEngine.Assertions;
 using UnityEngine.InputNew;
 using System.IO;
+using System.Reflection;
```
dunity: Sort usings

dunity:
I pushed a commit fixing the minor formatting changes

amirebrahimi:
@dunity - you requested changes to this, but then I think provided them, so you'll need to resolve those unless I've missed something that you still want me to change.

@mtschoen-unity - I've updated the branch and fixed the conflicts from latest development, so this should be good to go now.

dunity:
I pushed another two small commits, removing an unused using, and correcting a variable name.  In doing so I pushed one commit before pulling (my fetch looked up to date :-/ ).  Anyhow, the diff against @Amir's merge in of Development only show those two changes I mentioned, with no other differences.,

mtschoen-unity:
@amirebrahimi thanks for the update. To be honest, I haven't had a chance to look at this yet. Do you think this should go in before we merge #486?

amirebrahimi:
> @amirebrahimi thanks for the update. To be honest, I haven't had a chance to look at this yet. Do you think this should go in before we merge #486?

Looks like you already merged #486 to dev, which is why I updated this branch

Scripts/Core/Contexts/EditingContextManager.cs
```diff
@@ -47,13 +47,14 @@ internal static IEditingContext defaultContext
...
                 var context = availableContexts.Find(c => c.Equals(s_DefaultContext)) ?? availableContexts.First();
 
-                var defaultContextName = settings.defaultContextName;
-                if (!string.IsNullOrEmpty(defaultContextName))
-                {
-                    var foundContext = availableContexts.Find(c => c.name == defaultContextName);
-                    if (foundContext != null)
-                        context = foundContext;
-                }
+                // TODO: Fix context switching while EXR is running
```
amirebrahimi: How come? It doesn't always seem to screw up.

Scripts/Modules/WorkspaceModule.cs
```diff
@@ -148,6 +149,9 @@ public void OnDeserializePreferences(object obj)
                 var workspaceType = Type.GetType(workspaceLayout.name);
                 if (workspaceType != null)
                 {
+                    if (HiddenTypes.Contains(workspaceType))
```
amirebrahimi: Why this?

mtschoen-unity:
In testing this, I ran into some issues with changing contexts while EXR is running. The rays render to a large quad, and the cones are not visible. It seems that we've always supported switching contexts at runtime, but never had more than one context to test the swap. For now, I'm going to comment out the code which adds the context switcher UI to the VR view. For now, you will have to switch context by changing the default context asset reference on the `EditingContextManager.DefaultContext` asset reference on the EditingContextManager script importer, or setting the field on that script in a scene for play mode.

I also encountered the following issues:
- Builds fail due to changes in EditorVRContext. It seems like we just need to do the same thing we did for default tools--convert them to class names and rely on the name lists, not the script importers.
- The following tests were failing:
  - RestorePreviousContext_SetsPreviousContextToCurrent
  - SetEditingContext_SetsCurrentContext_IfNotNull
  - Utilities/MaterialUtilsTests
- Minor thing, but I removed the "?" on the preferences item for auto open because none of the other options had one
- Recovered/saved workspaces did not respect hidden types
- I can't get the spatial menu to come up in player builds--I don't think this needs to hold up the PR

Scripts/Modules/WorkspaceModule.cs
mtschoen-unity: I mention this in my comment below.  I found it odd that workspaces from a previous session, which won't be available in this particular context, come up when starting EXR

Scripts/Core/Contexts/EditingContextManager.cs
mtschoen-unity: In my testing, the rays would break every time. I see a big orange quad instead of the ray, and the cone disappears. I figured this was a separate concern from introducing minimal mode, so I didn't want to hold up the PR on fixing this.

mtschoen-unity: Aaaand now it's not breaking 🤔 
I will revert

mtschoen-unity: On further investigation, it seems only to break when going from EditorXR -> Minimal mode. I'll investigate this a bit more

amirebrahimi: Yeah, wasn't able to get to the bottom of this. We could force a whole restart of the VRView as a fix, but if you get to the root cause of it kudos to you.

mtschoen-unity: Actually, it doesn't seem to matter in what order I switch. It appears to happen every time I switch once the HMD and controllers are recognized. I did have a few successful switches earlier that I can't account for.

It appears that at least the controller issue is due to the fact that after switching the camera parent has a scale of 1000. I'm trying to find out where this happens.

mtschoen-unity: I think I've figured it out. It's not that switching causes the viewer scale to change, but if you switch with a non-zero viewer scale, the proxy ray shows up wrong. I'll see if I can find a fix.

PR #505, opened 2019-01-31
Refactor Create Primitive menu UI

dunity:
**Purpose of this PR**
Remove the automatic display of primitive shapes in the Create Primitive tool menu UI.  Re-design the UI to display the name of corresponding shape only (unless hovered, which displays the the shape).

**Testing status**
Tested thoroughly on both hands/proxies in both a clean scene, and a scene with existing content in it.

**Technical risk**
_Low._  The changes made _mostly_ affect only the CreatePrimitiveMenu constituent elements.
That aside, there were tangential bug fixes made in the MainMenuButton, NumericInputField, and RadialFrame shader in this PR.  There was a hover event/action issue fixed in the MainMenuButton, and a 2018.3 compatibility fix made to the NumericInputField.  I tested these fixes as well, and found no issues with them.  The RadialFrameShader had an unnecessary UV channel selection property removed.

**Comments to reviewers**
There were also hoverStarted & hoverEnded UnityEvents added to the MainMenu button.  No only have I had to constantly mod/patch these in on my own projects, but they are used by the buttons in the CreatePrimitiveMenu due to their not being bound by the (non-inspector exposed) hovered and clicked events that don't serve the same purpose due to their signatures (& no actual hoverEnd event/action).  These events allow for inspector-based subscription; which is also more convenient.  These new UnityEvents are raised in the _same_ cases as any existing actions in the class.

A core haptics (hover, click) pass was also performed on the Create Primitive menu's buttons.  The corresponding haptic-pulses used in the MainMenu (face) buttons were used in this case.

The "Close" button doesn't match the style of the other UI in the project, but was out of the scope of this UI _button_ redesign.  In addition a haptics pass should be made on the Primitive Menu.


amirebrahimi:
Looks good!

PR #509, opened 2019-02-10
Adjust resolution for high-DPI screens

amirebrahimi:
### Purpose of this PR

Minor fix for the profiler workspace

### Testing status

Manually tested

### Technical risk

Technical risk: low
Halo: low

### Comments to reviewers

Standard GUI is assuming a 96 dpi screen, so adjusting for DPIs that are higher than that (e.g. my laptop)


mtschoen-unity:
Thanks for this! I think this should also fix the console workspace.

PR #520, opened 2019-02-19
Menu/Ray hiding fix

mtschoen-unity:
### Purpose of this PR

Fix a bug introduced by #517 where the priority-1 ray visibility settings would override tools trying to hide the ray

### Testing status

Tested by opening Annotation and/or CreatePrimitive tool. Previously the ray would show on the primary tool hand. Now it does not. Also tested the ray returning after hiding the menu when in proximity to a workspace, which is what the priority-1 change was fixing.

### Technical risk

Low - Only affects ray visibility in a particular edge case

### Comments to reviewers

Good 'ol bug whack-a-mole


dunity:
Just tested by opening either tool, then closing the custom UI with the opposite hand.  However the ray doesn't appear again after closing the menu.  (Tested in the clean 2018.2 base project using this branch).

PR #522, opened 2019-02-19
Misc performance improvements

mtschoen-unity:
### Purpose of this PR

Introduce minor performance optimizations

### Testing status

Normal smoke test--open all workspaces and tools. This reduces about 1k of per-frame GC under normal circumstances, but a little more before the first selection is made (see below)

### Technical risk

Low--All code should be functionally equivalent, but a little faster

### Comments to reviewers

This was an old branch I opened last year but never got around to PR-ing

Note: includes `auto-open-init-fix` for convenience

What was done:
- Include inactive children while searching for manipulators. This avoids extra work done before the first selection is made searching for the manipulator in the EXR hierarchy
- The hierarchy and project workspaces now use int indices instead of strings, which should be a little faster when doing lookups in the dictionaries which contain them
- We were checking for each individual event type in `HoveringInteractable`, where a single check for all of them will do. This may pick up on some interaction types we haven't accounted for, but so far all of our UI is under our control, and there shouldn't be an issue with hovering over objects which are selectable but not clickable, for example.
- See Octree PR. Basically I replaced a class with a struct to avoid GC. https://github.com/Unity-Technologies/UnityOctree/pull/6


dunity:
Just a heads up before review: I just pushed a merge-in of latest dev.

mtschoen-unity:
> the radial menu now consumes the SpatialMenu thumbstick/trackpad input

Is this an issue on `development`?

> my first guess is the change to the MultiRayInputModule is the most likely culprit

Hm... maybe part of the SpatialMenu is registering as a valid hover target but shouldn't?

I'll see if I can reproduce the same issue.

PR #552, opened 2019-05-20
Two-Handed Rotation

mtschoen-unity:
### Purpose of this PR

Add a mode to two-handed manipulation to use both hands for rotation and scale simultaneously

### Testing status

Tested in Adventure Demo scene and with a basic cube

### Technical / Halo risk

Tech Risk 1 - could potentially introduce bugs to TransformTool

Halo Risk 0 - Only affects TransformTool and direct manipulation code

### Comments to reviewers

This is going up now so you guys have some time to look it over. It's pending a final design review.

The tricky part was figuring out how to get the "average" rotation change of the two hands. You can end up with some unstable situations when the axis vectors (forward of up on the rayorigin) face each other, but I found this to be preferable to what it looks like to use my first approach, which was to use `Quaternion.Lerp(leftHand, rightHand, 0.5f)`. That had a hard "pop" in similar situations.

The first pass allowed for two handed rotation-only as well as scale-only and scale-and-rotation. After a design review, we decided to punt on this until we can add some visuals to show what is going on during this manipulation. My hope is to land what we have here in order to close out this branch as a first pass.


AndrewTHEManeri:
Bug: Initiate a 2 handed scale + rotate, then release your second hand and regrab.
The tooltip will show the label for 'press to enable scale and rotate' but the manipulation is _already_ scale + rotate

mtschoen-unity:
@AndrewTHEManeri bug fixed. Good spotting 👍

mtschoen-unity:
>The snap at certain angles is not ideal but I think that's a good thing for our Design Engineer to figure out

Ugh yeah that was the big challenge here. I've found that it only "snaps" (I usually get it being really sensitive at the critical angle but not a hard snap) when you try to mess with it. It happens when you point the rotation rays (either forward or up depending on how you started the interaction) of each hand toward each other and their dot product approaches zero. I tried a bunch of different approaches that ended up having worse gimbal lock issues.

AndrewTHEManeri:
My intuition is that we actually want to ignore the hand-rotation and examine how the rotation of the axis formed by the two hands changes - probably paying attention to delta changes for directional consistency and then mapping back onto an absolute frame

mtschoen-unity:
The issue there is that you only get a 2D rotation. I think that might be interesting as a mode, as well, but this feels more powerful

PR #557, opened 2019-09-30
Staging 0.3 prep

amirebrahimi:
### Purpose of this PR

Fix compilation on 2017.4, so it at least shows download dialog box
Re-serialize assets for 2018.4, since that is the new min version
Update asmdef patch for pre-2019.1

### Testing status

Manual testing on 2017.4 and 2018.4

### Technical / Halo risk

[Overall product level assessment of the risk of the changes using the following criteria:

Tech Risk (not Tech Correctness, which is deferred to the reviewers)
* **0** could be just zero impact changes, removals but no logic changes
* ### **1** low impact, simple logic changes ###
* **2** is anything between 1 and 3
* **3** are extremely likely to introduce bugs

Halo Risk (Externality risk)
* **0** local change with no risk to other areas
* ### **1** neighbors could be affected ###
* **2** anything between 1 and 3
* **3** things might break everywhere]


Editor/BuildCallbacks.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if UNITY_2018_3_OR_NEWER
+﻿#if UNITY_2018_4_OR_NEWER
```
mtschoen-unity: I'd like to keep these as `2018_3_OR_NEWER` because, while we don't support all versions of 2018.3, 2019.1 doesn't include this define and this will cause compile issues

mtschoen-unity:
Thanks for updating the 2019_1_OR_NEWER stuff. I agree that ENABLE_EDITORXR is the way to go.

I'm finally getting around to testing this and having some trouble importing into an empty project. I'm still trying to figure out the best path forward, but after a clean import I get a bunch of null refs trying to start EditorXR. I'm guessing that it has to do with default references not getting set up properly if the first compile fails.

We can probably get rid of the TextMeshPro asmdef in the unitypackage patch thing, since 2018.4 includes the TMP package by default.

amirebrahimi:
TMP asmdef is to allow it to compile in versions prior to 2018.4 only to report that you should download 2018.4 :)

PR #560, opened 2019-11-01
Release fixes

mtschoen-unity:
### Purpose of this PR

Apply critical regression fixes.

### Testing status

Tested a variety of scene manipulation use cases like adding prefabs from project workspace, using preview camera, and duplicating/deleting/placing objects

### Technical / Halo risk

[Overall product level assessment of the risk of the changes using the following criteria:

Tech Risk (not Tech Correctness, which is deferred to the reviewers)
* **0** could be just zero impact changes, removals but no logic changes
* ### **1** low impact, simple logic changes
* **2** is anything between 1 and 3
* **3** are extremely likely to introduce bugs

Halo Risk (Externality risk)
* ### **0** local change with no risk to other areas
* **1** neighbors could be affected
* **2** anything between 1 and 3
* **3** things might break everywhere]

### Comments to reviewers
Some of these fixes appear to be due to changes in behavior in newer Editor versions--the related code was not changed since version 0.2.1.

Fixes:
* MonoScripts can fail to import either because the filenames don't match or because of first-time-import bugs (I was seeing one with `InspectorWorkspace`, which is fixed here). This results in `MonoScript.GetClass()` returning null, which was causing `NullReferenceExceptions` in `EditorVRContext.SetupMonoScriptTypeNames`
* InspectorWorkspace didn't import properly because its "stub" class does not exist when `ENABLE_EDITORXR` isn't defined on first import
* An exception can occur during shutdown when trying to access `m_PlayerBody.gameObject` in `EditorVR.Viewer.cs`
* `InputTracking.GetLocalPosition` is marked as deprecated in 2019.2, but unfortunately the newer API which replaces it does not update in edit mode. I have brought this up with the XR Features team.
* All versions exhibited an odd behavior where the UI content inside the VR view scaled weirdly based on window size when using the presentation camera. It seems like the GUI matrix is being transformed somehow because the context dropdown and device view/presentation camera toggles also get stretched and squished. The fix is to ensure that `XRSettings.showDeviceView` is true when the presentation camera is in use, as well as the regular device view. The same issue occurs if the device view is disabled and we continue to call `Handles.DrawCamera`. This appears to be new behavior introduced in 2018.4 and above.
* In 2019.1 and above, the regular device view (no presentation camera) is flipped on the Y-axis, so there is now some version-specific code to flip the texture for the regular device view in 2019.1 and avove.
* The `BoundsExtensions.ContainsCompletely` method was giving false positives when bounding boxes aligned in one axis but did not overlap in another, causing objects to be not-selectable when the user was not contained within them.
* The `VRSmoothCamera` was creating its render texture every frame because it was checking if its render texture was the same size as the VR  camera's, but then scaling that rectangle before creating its own render texture.
* The Project and Poly Workspaces didn't re-arrange the grid when scaling the icons (due to list view update)

I've also noticed an intermittent bug where haptics seem to get "stuck on," but I haven't been able to reproduce this reliably.

### DIFF
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -436,6 +436,11 @@ namespace UnityEditor.Experimental.EditorVR.Core
             {
                 if (e.type == EventType.Repaint)
                 {
+#if UNITY_2019_1_OR_NEWER
+                    if (!customPreviewCamera)
+                        guiRect = new Rect(guiRect.position.x, guiRect.position.y + guiRect.height, guiRect.width, -guiRect.height);
+#endif
+
                     var renderTexture = customPreviewCamera && customPreviewCamera.targetTexture ? customPreviewCamera.targetTexture : m_TargetTexture;
                     GUI.DrawTexture(guiRect, renderTexture, ScaleMode.StretchToFill, false);
                 }
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -464,6 +464,9 @@ namespace UnityEditor.Experimental.EditorVR.Core
             if (!XRDevice.isPresent)
                 return;
 
+            if (!m_ShowDeviceView)
+                return;
+
             UnityEditor.Handles.DrawCamera(rect, m_Camera, m_RenderMode);
             if (Event.current.type == EventType.Repaint)
             {
--- a/Scripts/Core/VRView.cs
+++ b/Scripts/Core/VRView.cs
@@ -371,7 +371,9 @@ namespace UnityEditor.Experimental.EditorVR.Core
             // Always render camera into a RT
             CreateCameraTargetTexture(ref m_TargetTexture, cameraRect, false);
             m_Camera.targetTexture = m_TargetTexture;
-            XRSettings.showDeviceView = m_ShowDeviceView;
+            //XRSettings.showDeviceView = m_ShowDeviceView;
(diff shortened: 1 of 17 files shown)


Scripts/Extensions/BoundsExtensions.cs
```diff
@@ -10,11 +10,8 @@ static class BoundsExtensions
         /// <param name="otherBounds">The bounds to compare with this one</param>
         public static bool ContainsCompletely(this Bounds bounds, Bounds otherBounds)
         {
-            if (bounds.min.MinComponent() > otherBounds.min.MinComponent()
-                || bounds.max.MinComponent() < otherBounds.max.MinComponent())
-                return false;
-
-            return true;
+            return bounds.max.x >= otherBounds.max.x && bounds.max.y >= otherBounds.max.y && bounds.max.z >= otherBounds.max.z
+                   && bounds.min.x <= otherBounds.min.x && bounds.min.y <= otherBounds.min.y && bounds.min.z <= otherBounds.min.z;
```
mtschoen-unity: Cache bounds.min, etc

Scripts/Core/VRView.cs
```diff
@@ -464,6 +471,9 @@ void DoDrawCamera(Rect rect)
             if (!XRDevice.isPresent)
                 return;
 
+            if (!m_ShowDeviceView)
+                return;
+
```
mtschoen-unity: Test this

